Issue #10, opened 2020-08-21
Expand the set of sample bots


Nihlus:
Currently, a single example bot is available under the samples directory, which simply connects to the gateway and logs unknown or unparseable events to file.

More samples that demonstrate various common tasks (such as sending messages, creating embeds, modifying roles and permissions, etc) should be developed as part of the 1.0 release and typical usage documentation.

PR #20, opened 2020-10-06
Add HttpCat sample.

Tylertron1998:
This adds a httpcat sample, using the free http.cat api - as per #10.


Nihlus:
Looks good :+1:

PR #22, opened 2020-11-07
Added a simple Microsoft.Extensions.Hosting implementation

Nikey646:
Created as a new project to not require the abstractions for users who don't want to use Microsoft.Extensions.Hosting.

(P.S: Sorry about accidentally including the riderModule.iml, and I'm a git nub so don't know how to flatten the commits)


Remora.Discord.Hosting/Services/DiscordService.cs
```diff
@@ -0,0 +1,77 @@
...
+            this._gatewayClient = gatewayClient;
+            this._logger = logger;
+        }
+
+        /// <inheritdoc />
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            while (!stoppingToken.IsCancellationRequested)
+            {
+                var runResult = await this._gatewayClient.RunAsync(stoppingToken);
```
Nihlus, line 56: Generally, the gateway client's `RunAsync` method shouldn't return unless a fatal, non-recoverable error has occurred, or cancellation has been requested via the token. Keeping it alive in a loop like this should not be required.

When I say non-recoverable, I mean in the class of things like "your token is invalid" or "there's a bug that's throwing and it's going to happen again". There *might* be a case where the internet connection isn't available and because of that it returns, and then a loop like this might be warranted - I'm unsure at the time of writing whether that's implicitly handled somewhere in the inner chain, though. Mental note to check.

Nihlus (changes requested):
Thanks! All in all, this looks pretty good, minimalistic as it is. I really should fix the header requirement so that it doesn't force you to use my name as the author, though. Copyright assignment should still be there, though, so I'll have to think about it for a little while.

There are some changes that you should take a look at, but they're minimal.

Remora.Discord.Hosting/Remora.Discord.Hosting.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <PropertyGroup>
+        <TargetFramework>netstandard2.1</TargetFramework>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="..\Backend\Remora.Discord.Gateway\Remora.Discord.Gateway.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="5.0.0-rc.2.20475.5" />
```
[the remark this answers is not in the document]
Nikey646: My bad, I let Rider automatically select the version I think and it must have grabbed the Pre Release.

Remora.Discord.Hosting/Services/DiscordService.cs
Nikey646: Should I then make the loop only continue if there is a value in In the Websocket / Gateway close status?

I want to avoid any chance of the bot seemingly stopping for no reason, hence why the 5 second delay and attempting to reconnect.

Nihlus: That's quite reasonable :) While it shouldn't be neccesary, it's not a bad precaution.

PR #24, opened 2020-11-08
Add support to load all Responders in an Assembly.

Nikey646:
The objective of this PR is to add a user friendly way to load all types implementing `IResponder<T>` in an assembly. With the current implementation, it is not possible to do that, since `ResponderService.RegisterResponder<TResponder>()` is not only internal but does not provide a non-generic overload.

### DIFF
--- a/Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
+++ b/Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
@@ -80,31 +80,6 @@ namespace Remora.Discord.Gateway.Extensions
         }
 
         /// <summary>
-        /// Adds all exported responders in the provided assembly to the service collection.
-        /// This method registers the responder as being available for all <see cref="IResponder{T}"/>
-        /// implementations it supports.
-        /// </summary>
-        /// <param name="serviceCollection">The service collection.</param>
-        /// <param name="assembly">The assembly to search.</param>
-        /// <returns>The service collection, with the responder added.</returns>
-        public static IServiceCollection AddRespondersFromAssembly
-        (
-            this IServiceCollection serviceCollection,
-            Assembly assembly
-        )
-        {
-            var types = assembly.GetExportedTypes()
-                .Where(t => t.IsResponder());
-
-            foreach (var type in types)
-            {
-                serviceCollection.AddResponder(type);
-            }
-
-            return serviceCollection;
-        }
-
-        /// <summary>
         /// Adds a responder to the service collection. This method registers the responder as being available for all
         /// <see cref="IResponder{T}"/> implementations it supports.
         /// </summary>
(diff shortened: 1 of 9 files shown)


Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -75,22 +76,69 @@ Func<string> token
...
+        /// <param name="assembly">The assembly to search.</param>
+        /// <returns>The service collection, with the responder added.</returns>
+        public static IServiceCollection AddRespondersFromAssembly
+        (
+            this IServiceCollection serviceCollection,
+            Assembly assembly
+        )
+        {
+            // ReSharper disable once SuspiciousTypeConversion.Global
+            var types = assembly.GetExportedTypes()
```
Nihlus, line 97: This won't work, since `t` will always be an instance of `Type`, which does not implement `IResponder`. You'll need to check the interfaces implemented by the type instead.

Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -75,22 +76,69 @@ Func<string> token
...
+        /// <see cref="IResponder{T}"/> implementations it supports.
+        /// </summary>
+        /// <param name="serviceCollection">The service collection.</param>
+        /// <param name="responderType">The type implementing <see cref="IResponder"/>.</param>
+        /// <returns>The service collection, with the responder added.</returns>
+        /// <exception cref="ArgumentException">Throws if responderType does not implement <see cref="IResponder"/>.</exception>
+        public static IServiceCollection AddResponder(this IServiceCollection serviceCollection, Type responderType)
+        {
+            // ReSharper disable once SuspiciousTypeConversion.Global
+            if (!(responderType is IResponder))
```
Nihlus, line 119: Same issue here. Have a look at how the existing `AddResponder` method does it.

Backend/Remora.Discord.Gateway/Services/ResponderService.cs
```diff
@@ -43,7 +43,24 @@ public class ResponderService : IResponderTypeRepository
...
+        }
+
+        /// <summary>
+        /// Adds a responder to the service.
+        /// </summary>
+        /// <param name="responderType">The responder type.</param>
+        internal void RegisterResponderType(Type responderType)
+        {
+            // ReSharper disable once SuspiciousTypeConversion.Global
+            if (!(responderType is IResponder))
```
Nihlus, line 56: And here as well.

Nihlus (changes requested):
This is a pretty good addition, but the way you've implemented it won't work. Make sure you've ran all unit tests locally before pushing.

Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
Nikey646: ... I must not have been thinking when I wrote that code. I'll fix that up in a few moments.

PR #25, opened 2020-12-20
Fix file uploads

wtfblub:
File uploads require a filename or the upload will fail. This change introduces a way for users to specify a filename for the given content stream.

Doc reference: https://discord.com/developers/docs/resources/channel#create-message see the extra note.

> For a file attachment, the Content-Disposition subpart header MUST contain a filename parameter.


Backend/Remora.Discord.API/API/Objects/Messages/File.cs
```diff
@@ -0,0 +1,38 @@
...
+using JetBrains.Annotations;
+using Remora.Discord.API.Abstractions.Objects;
+
+#pragma warning disable CS1591
+
+namespace Remora.Discord.API.Objects
+{
+    /// <inheritdoc cref="IFile" />
+    [PublicAPI]
+    public record File
```
Nihlus, line 33: See the comment on `IFile` - considering this type doesn't map to a Discord type, having an interface might even be overkill. Thoughts?

Tests/Remora.Discord.Rest.Tests/API/Channels/DiscordRestChannelAPITests.cs
```diff
@@ -745,7 +752,7 @@ public async Task PerformsFileUploadRequestCorrectly()
                     channelId,
                     nonce: nonce,
                     isTTS: tts,
-                    file: file
+                    file: new Discord.API.Objects.File(fileName, file)
```
Nihlus, line 755: Import this namespace to avoid full qualification.

Tests/Remora.Discord.Rest.Tests/API/Webhooks/DiscordRestWebhookAPITests.cs
```diff
@@ -780,7 +787,7 @@ public async Task PerformsFileUploadRequestCorrectly()
                     token,
                     shouldWait,
                     isTTS: tts,
-                    file: file
+                    file: new Discord.API.Objects.File(fileName, file)
```
Nihlus, line 790: See previous comment.

Nihlus (changes requested):
Thanks, good catch! I do have some requests before merging it in.

wtfblub:
Renamed the File class to FileData to avoid System.IO.File name conflicts and moved it to the Remora.Discord.Core library.

Nihlus:
Perfect, thanks :)

PR #27, opened 2020-12-29
Fix exception on multiple command options

wtfblub:
Fixes an exception when using multiple options on a slash command. `SingleOrDefault` throws an exception when theres multiple options so the logic here was slightly wrong.

### DIFF
--- a/Remora.Discord.Commands/Responders/InteractionResponder.cs
+++ b/Remora.Discord.Commands/Responders/InteractionResponder.cs
@@ -163,8 +163,7 @@ namespace Remora.Discord.Commands.Responders
             commandStringBuilder ??= new StringBuilder();
             parameters ??= new Dictionary<string, IReadOnlyList<string>>();
 
-            var singleOption = options.SingleOrDefault();
-            if (singleOption is null)
+            if (options.Count > 1)
             {
                 // multiple parameters
                 foreach (var option in options)
@@ -174,6 +173,7 @@ namespace Remora.Discord.Commands.Responders
             }
             else
             {
+                var singleOption = options.Single();
                 if (singleOption.Value.HasValue)
                 {
                     // A single parameter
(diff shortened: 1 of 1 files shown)


Remora.Discord.Commands/Responders/InteractionResponder.cs
```diff
@@ -174,6 +173,7 @@ IApplicationCommandInteractionData commandData
             }
             else
             {
+                var singleOption = options.Single();
```
Nihlus, line 176: This would miss parameterless commands - use SingleOrDefault and check for null.

Nihlus (changes requested):
Ah, damn. Thanks! We really should get some unit tests set up for the unpacking logic.

Remora.Discord.Commands/Responders/InteractionResponder.cs
wtfblub: Parameterless commands have worked for me but I still added a check just in case.

Nihlus: This wasn't exactly what I meant - now, if there are no options, all branches will be missed. I'm interested in the fact that parameterless commands were working for you - to me, it shouldn't when you look at this branching. I'll merge this, but won't push it live until I've implemented some unit tests and figured out what the hell I've actually written here x)

wtfblub: Parameterless commands probably hit this path
https://github.com/wtfblub/Remora.Discord/blob/72e2209f130aa419187279a27a1ae45595387e32/Remora.Discord.Commands/Responders/InteractionResponder.cs#L141-L144

Nihlus: Quite right - I've refactored the paths a bit and added unit tests for all cases, just to be on the safe side. Fixed version should be up on nuget soon.

You should add yourself to the AUTHORS file, by the way :P

PR #33, opened 2021-01-10
Allow for the bot token to be supplied via Dependency Injection

### LINKED ISSUE #32: Allow for the bot token to be supplied via Dependency Injection
**Description**

Currently, the bot token is supplied to the `.AddDiscordGateway()` extension method via a `Func<string>` delegate, which accepts no parameters, and effectively means that the token must be available in scope during the `ConfigureServices` phase of application startup.

Instead, it should be possible to leverage the Dependency Injection system to retrieve the token, such that it is not required until the IoC container has been constructed.

I.E. the token factory delegate should be of type `Func<IServiceProvider, string>()`

**Why This is Needed**

Allowing the token to be retrieved after the IoC container has been constructed allows configuration binding and validation systems, such as `Microsoft.Extensions.Options` to be leveraged to retrieve the token.

For example, a token factory delegate could be...

```cs
serviceProvider => serviceProvider.GetRequiredService<IOptions<MyBotOptions>>()
    .Value
    .BotToken
```

**Alternatives Considered**

As it currently stands, retrieving the bot token is possible by manually extracting it from a `Microsoft.Extensions.Configuration` system.

For example, the token factory delegate that would work today, and is equivalent to the one given above would be...

```cs
() => configuration.GetSection("MyBot")["BotToken"]
```

This is undesirable, since it would require manual validation of the token, if desired, and would throw non-meanginful exceptions if the config data is missing or malformed.

**Additional Details**

From what I can tell, the library is ALREADY setup for injecting the token via IoC, but the IoC container is simply hidden, as seen [here](https://github.com/Nihlus/Remora.Discord/blob/57712e7b7fe7e2bb4e3eba299c5da9f3ca57064a/Backend/Remora.Discord.Rest/Extensions/ServiceCollectionExtensions.cs#L125)

```cs
.AddSingleton<ITokenStore>(_ => new TokenStore(token()));
```

An `IServiceProvider` is already available here, it's just not passed on to the token factory delegate, and there's no reason it can't be..

```cs
.AddSingleton<ITokenStore>(serviceProvider => new TokenStore(token(serviceProvider)));
```

JakenVeina:
Forward an IoC container to the factory methods for configuring bot tokens, to allow the token to be retrieved from the DI system.

Fixes #32.


JakenVeina:
The build seems to have failed because a test timed out. All tests run fine for me, locally.

PR #40, opened 2021-01-24
Reworked `IPayload<T>` to not allow for null data fields

JakenVeina:
This eliminates the need for responders to allow for nullable data fields when the vast majority of them can never be null.

### DIFF
--- a/Backend/Remora.Discord.API/Json/Converters/Internal/PayloadConverter.cs
+++ b/Backend/Remora.Discord.API/Json/Converters/Internal/PayloadConverter.cs
@@ -28,6 +28,7 @@ using Remora.Discord.API.Abstractions.Gateway;
 using Remora.Discord.API.Abstractions.Gateway.Bidirectional;
 using Remora.Discord.API.Abstractions.Gateway.Commands;
 using Remora.Discord.API.Abstractions.Gateway.Events;
+using Remora.Discord.API.Gateway.Bidirectional;
 using Remora.Discord.API.Gateway.Events;
 using Remora.Results;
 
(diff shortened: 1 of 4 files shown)


Backend/Remora.Discord.API/Json/Converters/Internal/PayloadConverter.cs
```diff
@@ -252,13 +253,44 @@ private RetrieveEntityResult<OperationCode> GetOperationCode(Type objectType)
...
+
+            if (data is null)
+            {
+                throw new JsonException();
+            }
+
             return new Payload<TData>(data);
         }
 
+        private static IPayload DeserializeHeartbeatPayload(JsonElement dataProperty, JsonSerializerOptions options)
```
Nihlus, line 269: This method should not be neccesary; see https://github.com/Nihlus/Remora.Discord/blob/master/Backend/Remora.Discord.API/Json/Converters/Internal/HeartbeatConverter.cs

Backend/Remora.Discord.API/Json/Converters/Internal/PayloadConverter.cs
```diff
@@ -252,13 +253,44 @@ private RetrieveEntityResult<OperationCode> GetOperationCode(Type objectType)
...
+        {
+            var data = JsonSerializer.Deserialize<long?>(dataProperty.GetRawText(), options);
+
+            return new Payload<IHeartbeat>(new Heartbeat(data));
+        }
+
+        private static IPayload DeserializeInvalidSessionPayload(JsonElement dataProperty, JsonSerializerOptions options)
```
Nihlus, line 276: This method should not be neccesary; see https://github.com/Nihlus/Remora.Discord/blob/master/Backend/Remora.Discord.API/Json/Converters/Internal/InvalidSessionConverter.cs

Nihlus (changes requested):

Backend/Remora.Discord.API/Json/Converters/Internal/PayloadConverter.cs
JakenVeina: Indeed.

Backend/Remora.Discord.API/Json/Converters/Internal/PayloadConverter.cs
JakenVeina: Indeed.

PR #45, opened 2021-02-05
JSON Sample Data Test Naming

JakenVeina:
Enhanced the JSON Sample Data testing system to display more meaingful test names, I.E. to have test names only include the relative path of the JSON file, not the full filesystem path, which would generally result in test names being truncated and unreadable within an IDE.

### DIFF
--- a/Tests/Remora.Discord.API.Tests/Services/SampleDataDescriptor.cs
+++ b/Tests/Remora.Discord.API.Tests/Services/SampleDataDescriptor.cs
@@ -1,0 +1,65 @@
+﻿//
+//  SampleDataDescriptor.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) 2017 Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


Tests/Remora.Discord.API.Tests/Services/SampleDataDescriptor.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>
+        /// Gets the full filesystem path of the sample data file.
+        /// </summary>
+        public string FullPath
+            => Path.Combine(BasePath, SubPath);
+
+        /// <summary>
+        /// Gets the path to the sample data file, relative to other sample data files.
+        /// </summary>
+        public string SubPath { get; }
```
Nihlus, line 59: This should be named RelativePath in order to keep things consistent.

Nihlus (changes requested):

Tests/Remora.Discord.API.Tests/Services/SampleDataDescriptor.cs
JakenVeina: Fair point.

Nihlus (approved):

PR #51, opened 2021-02-16
Add a condition for checking if the user has a given permission in the guild in which the command was executed.

TraceLD:
Added a condition for checking if the user has a given permission in the guild in which the command was executed. This is very useful for commands such as `!ban` etc. Also added an error that is used when a condition is not met.

### DIFF
--- a/Remora.Discord.Commands/Conditions/Attributes/RequireUserGuildPermissionAttribute.cs
+++ b/Remora.Discord.Commands/Conditions/Attributes/RequireUserGuildPermissionAttribute.cs
@@ -41,7 +41,7 @@ namespace Remora.Discord.Commands.Conditions
         /// <param name="permission">The permission.</param>
         public RequireUserGuildPermissionAttribute(DiscordPermission permission)
         {
-            Permission = permission;
+           this.Permission = permission;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Remora.Discord.Commands/Conditions/Attributes/RequireUserGuildPermissionAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// </summary>
+        public DiscordPermission Permission { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RequireUserGuildPermissionAttribute"/> class.
+        /// </summary>
+        /// <param name="permission">The permission.</param>
+        public RequireUserGuildPermissionAttribute(DiscordPermission permission)
+        {
+            Permission = permission;
```
Nihlus, line 44: Prefix public member fields and properties with `this.`

Remora.Discord.Commands/Conditions/RequireUserGuildPermissionCondition.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            var getGuildRoles = await _guildApi.GetGuildRolesAsync(guildId, ct);
+            if (!getGuildRoles.IsSuccess)
+            {
+                return Result.FromError(getGuildRoles);
+            }
+
+            var guildRoles = getGuildRoles.Entity;
+            var everyoneRole = guildRoles.FirstOrDefault(r => r.Name.Equals("@everyone"));
```
Nihlus, line 89: You should probably check for `guild.ID == role.ID` instead - it's a bit more robust in this case.

Nihlus (changes requested):

TraceLD:
I have applied the suggestions from your code review @Nihlus.

PR #57, opened 2021-04-09
Remove trailing "?" when there are no query params

thaumanovic:
Small change to remove the trailing question mark in query strings when there isn't actually a query. 

Unsure what the HTTP spec specifies in this case, but I don't feel it's semantically correct to leave it at the end when it's not being used.


Nihlus:
I'll see where this appears - I suspect it's something under the hood of the HttpClient that adds it.

PR #58, opened 2021-04-22
Redis Cache

GerardSmit:
This PR adds support for caching data in Redis.

## Motivation
Currently Remora only supports memory cache. This is fine if the bot handles the same shard every time. However when using Remora as a worker that handles multiple shards, the cache provider will miss a lot of cached values because they are stored in a different worker.

For example:
```
                  -- Gateway -- +-------------+ -- Worker (Remora.Discord)
                /               |             |                  
   Discord  -- + --- Gateway -- |  Rabbit MQ  | -- Worker (Remora.Discord)
                \               |             |                  
                  -- Gateway -- +-------------+ -- Worker (Remora.Discord)
```

To resolve this, a shared cache client will be used (Redis)

## Cache client
To support multiple cache clients (memory and Redis) a new interface called `ICacheClient` has been created:

```cs
public interface ICacheClient
{
    /// <summary>
    /// Stores the value in the cache.
    /// </summary>
    /// <param name="key">Key of the item.</param>
    /// <param name="value">Value to cache.</param>
    /// <typeparam name="TInstance">Type of the item.</typeparam>
    /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
    Task StoreAsync<TInstance>(CacheKey key, TInstance value);

    /// <summary>
    /// Attempts to retrieve a value from the cache.
    /// </summary>
    /// <typeparam name="TInstance">The instance type.</typeparam>
    /// <returns>Value from the cache provider.</returns>
    public Task<CacheResult<TInstance>> RetrieveAsync<TInstance>(CacheKey key) where TInstance : notnull;

    /// <summary>
    /// Removes value of the given key from the cache.
    /// </summary>
    /// <param name="key">Key of the value.</param>
    /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
    Task EvictAsync(CacheKey key);
}
```

This cache client is used in the cache service, instead of the currently used `IMemoryCache`.

## Cache Key
A new struct called `CacheKey` has been created which stores all the parameters of the key. When using Redis, these parameters will be converted into a single string (for example `IGuild:1234`). When using memory cache, the hash code will be used.

## Configuring
The user can install `Remora.Discord.Caching` for the current memory implementation:
```cs
// Use Memory
serviceCollection.AddDiscordCaching();
```

To use Redis (or any other distributed cache implemetation), the user has to install `Microsoft.Extensions.Caching.StackExchangeRedis` and call `CacheBuilder.UseDistributedCache()`:
```cs
// Use Redis
serviceCollection.AddStackExchangeRedisCache(options =>
{
    options.Configuration = connectionString;
});

serviceCollection.AddDiscordCaching(b => b.UseDistributedCache());
```


Backend/Remora.Discord.Caching.Redis/RedisCacheClient.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            TimeSpan? expiration;
+            DateTimeOffset? startSlidingAt;
+
+            if (expirationSetting != TimeSpan.MaxValue)
+            {
+                expiration = expirationSetting;
+                startSlidingAt = DateTimeOffset.Now.Add(expirationSetting).Subtract(slidingSetting);
+            }
+            else
```
Nihlus, line 76: You can omit this branch by simply initializing the variables at declaration time, which is generally preferred.

Backend/Remora.Discord.Caching/CacheKey.cs
```diff
@@ -0,0 +1,179 @@
...
+using System;
+using System.Diagnostics;
+
+namespace Remora.Discord.Caching
+{
+    /// <summary>
+    /// Represents a cache key.
+    /// </summary>
+    [DebuggerDisplay("{" + nameof(Key) + ",nq}")]
+    public readonly struct CacheKey : IEquatable<CacheKey>
```
Nihlus, line 32: This structure has a relatively heavy maintenance impact, considering that the argument count may grow in the future. See if you can generalize it by using a list or preallocated array (a `params` constructor, perhaps?).

Backend/Remora.Discord.Caching/CacheResult.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace Remora.Discord.Caching
+{
+    /// <summary>
+    /// Result from the cache.
+    /// </summary>
+    /// <typeparam name="T">Type of the value.</typeparam>
+    public readonly struct CacheResult<T> where T : notnull
```
Nihlus, line 31: This type is redundant; everything it does is available through `Result<T>`. Use that type instead, if there are no particular reasons for adding this type.

Backend/Remora.Discord.Caching/Extensions/CacheBuilder.cs
```diff
@@ -0,0 +1,46 @@
...
+//
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Remora.Discord.Caching.Extensions
+{
+    /// <summary>
+    /// Utility class used by <see cref="ServiceCollectionExtensions.AddDiscordCaching"/> to configure the cache.
+    /// </summary>
+    public class CacheBuilder
```
Nihlus, line 30: Any particular reason for this type? The one place it's being used could be replaced with an `Action<IServiceCollection>`, as far as I can tell - either that, or omitting it completely, simply letting the user call `UseMemory` or `UseRedis`, and altering those extension methods to add caching on their own. That way, you also avoid a user accidentally not registering any caching backend (if that route is taken, the methods should probably be renamed, too).

Backend/Remora.Discord.Caching/Services/CacheSettings.cs
```diff
@@ -53,6 +52,42 @@ public class CacheSettings
         /// </summary>
         private TimeSpan _defaultSlidingExpiration = TimeSpan.FromSeconds(10);
 
+        /// <summary>
+        /// Gets a value indicating whether the messages should be cached.
```
Nihlus, line 56: Change "the messages" to "messages".

Backend/Remora.Discord.Caching/Services/CacheSettings.cs
```diff
@@ -53,6 +52,42 @@ public class CacheSettings
...
+        /// </summary>
+        public bool ShouldCacheMessages { get; private set; }
+
+        /// <summary>
+        /// Gets a value indicating whether the presences should be cached.
```
Nihlus, line 61: Change "the presences" to "user presence information".

Backend/Remora.Discord.Caching/Services/CacheSettings.cs
```diff
@@ -53,6 +52,42 @@ public class CacheSettings
...
+        /// <summary>
+        /// Enables message cache.
+        /// </summary>
+        /// <remarks>
+        /// Since large bots will receive a lot of messages this should not be enabled unless necessary.
+        /// </remarks>
+        /// <returns>The settings.</returns>
+        public CacheSettings EnableMessageCache()
+        {
+            ShouldCacheMessages = true;
```
Nihlus, line 74: Properties should be qualified with `this.`.

Backend/Remora.Discord.Caching/Services/CacheSettings.cs
```diff
@@ -53,6 +52,42 @@ public class CacheSettings
...
+        /// <summary>
+        /// Enables presence cache.
+        /// </summary>
+        /// <remarks>
+        /// Since large bots will receive a lot of presence this should not be enabled unless necessary.
+        /// </remarks>
+        /// <returns>The settings.</returns>
+        public CacheSettings EnablePresenceCache()
+        {
+            ShouldCacheMessages = true;
```
Nihlus, line 87: Properties should be qualified with `this.`.

Backend/Remora.Discord.Caching/Extensions/CacheBuilder.cs
Nihlus: One option is also to fall back to memory caching if no other strategy has been configured. Perhaps an argument for retaining the memory cache in the main Caching package?

Nihlus (changes requested):
Thanks a ton for this! I've not used Redis myself, but this is a very welcome and well-written 
addition. I have some changes I'd like you to consider, as well as some questions regarding various changes.

Primarily, I'd like to avoid a situation where a user might install the caching package (or upgrade from the existing one) and have their bot fail because they forgot to explicitly choose a caching strategy. Preferably, the main caching package would still be functional as-is, and the Redis backend would be an optional alternative that could replace the default functionality. It's just a thought, so feel free to counter it :)

Backend/Remora.Discord.Caching/Extensions/CacheBuilder.cs
GerardSmit: > Any particular reason for this type? The one place it's being used could be replaced with an `Action<IServiceCollection>`

The only downside is that you'll get all the extension methods of the interface `IServiceCollection`. For example you could do the following:
```cs
services.AddDiscordCaching(x => x.AddDiscordCaching(y => y.AddDiscordCaching())); // just to make sure it gets registered 😄
```

With this new type, you'll only get the extensions methods that targets the cache building.
```cs
services.AddDiscordCaching(b => b.UseRedis());
```

> simply letting the user call UseMemory or UseRedis, and altering those extension methods to add caching on their own

From my understanding, you want to have a new extensions method called for example `services.AddDiscordRedisCaching()`?

> One option is also to fall back to memory caching if no other strategy has been configured.

I've added this. When no cache client is registered, the memory cache client will be added.

Noteworthy is that the Memory Cache will always be registered (even when Redis is being used) but maybe this is for the best, otherwise it would be a breaking change (if users depend on this without calling AddMemoryCache because AddDiscordCaching already does this).

Backend/Remora.Discord.Caching/CacheKey.cs
GerardSmit: I can change this to an array, however the only downside is that every time a cache key is being constructed, a new array will be allocated. Next to that, you'll have to loop through the array to make a hash code.

GerardSmit:
Oops. I forgot to push one commit causing the build to fail 😅

I've added questions in 2 requested changes. I've resolved the rest of them.

Nikey646:
@GerardSmit Is there a benefit to creating an entirely separate caching interface / solution for this PR, instead of implementing [`IDistributedCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.distributed.idistributedcache), which has support for In-Memory and Redis back ends built in, along side other back ends (Both from Microsoft directly and the community).

The only downside to implementing `IDistributedCache` that I can see if the usage of the it requires all data to be transformed to a `byte[]` before being stored, but looking at your Redis implementation in this PR, that is being done (But not for the In-Memory implementation), but that can be worked around quite easily.

Backend/Remora.Discord.Caching/Extensions/CacheBuilder.cs
Nihlus: Alright - looks good now :)

Backend/Remora.Discord.Caching/CacheKey.cs
Nihlus: Thinking outside the box for a moment, is there a better way of creating a cache key?

GerardSmit:
> @GerardSmit Is there a benefit to creating an entirely separate caching interface / solution for this PR, instead of implementing [`IDistributedCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.distributed.idistributedcache), which has support for In-Memory and Redis back ends built in, along side other back ends (Both from Microsoft directly and the community).

Because I didn't know that distributed cache had sliding cache expiration. I've updated the PR that it uses distributed cache.

> The only downside to implementing `IDistributedCache` that I can see if the usage of the it requires all data to be transformed to a `byte[]` before being stored, but looking at your Redis implementation in this PR, that is being done (But not for the In-Memory implementation), but that can be worked around quite easily.

I left the default implementation intact, so it isn't a breaking change. If we want to replace the `MemoryCache` to `DistributedMemoryCache` we should do this in a new major version.

Backend/Remora.Discord.Caching/CacheKey.cs
GerardSmit: I've added a new constructor that accepts an object array. When using less than 3 keys it doesn't create an object array and it uses the arg1..3 field (https://github.com/Nihlus/Remora.Discord/pull/58/commits/a514688a0f8d2e4c8c937e145b00d8be0b0296d0).

> Thinking outside the box for a moment, is there a better way of creating a cache key?

We could also just use a string in the `MemoryCache`-implemetation, instead of relying on the hash code.

Nihlus: I'd prefer not to have a hybrid solution; that's unneeded complexity. Choose one or the other.

String allocations should be avoided where possible, so relying on the hash code is preferable in that case.

Backend/Remora.Discord.Caching/Extensions/CacheBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <typeparam name="TClient">New cache client to use.</typeparam>
+        /// <returns>Same instance of the builder.</returns>
+        public CacheBuilder UseClient<TClient>() where TClient : class, ICacheClient
+        {
+            Services.AddSingleton<ICacheClient, TClient>();
+            return this;
+        }
+
+        /// <summary>
+        /// Replaces the default memory cache client with the distributed cache client.
```
Nihlus, line 62: This method doesn't actually replace anything; it just adds another more prioritized service. Either correct the comment or the implementation.

Nihlus (changes requested):

Backend/Remora.Discord.Caching/CacheKey.cs
GerardSmit: Using a string as key in Redis is very useful to look up the values:

image

Also, this way you can easily remove cache (for instance, removing all values with the pattern `IGuild:*`).
So honestly I would like to have a string as key in Redis.

Backend/Remora.Discord.Caching/Extensions/CacheBuilder.cs
GerardSmit: Changed, is this okay?

Backend/Remora.Discord.Caching/CacheKey.cs
Nihlus: For Redis, that's totally fine; I was more referring to the memory cache that currently uses the hashcode (and I think should continue to do so).

Backend/Remora.Discord.Caching/Extensions/CacheBuilder.cs
Nihlus: Yep, looks fine :)

Backend/Remora.Discord.Caching/CacheKey.cs
GerardSmit: It does... That's why I made the CacheKey. So I'm unsure what has to be done here now.

Nihlus: It's two things.

  1) Either remove the arg1, arg2, ... fields and just use the array, or remove the array and just use the arg1, arg2 fields. I would prefer that you did the former.
  2) Don't do anything to the keying - redis using strings and memory cache using the struct directly is perfectly fine.

TraceLD:
What will happen with the current design, if the Redis cache is accessed remotely (e.g. the Redis cache DB is hosted on a different server than the bot - e.g. bot on OVH VPS, cache on Azure) and the remote DB goes down? Will it bring down the entire bot or log the cache failure and just pull from Discord API?

Nihlus:
Any progress on this? :) I'd very much like to merge it, and it's not far from being complete. It could use some rebasing now that the v9 API is implemented, though.

Nihlus:
Closing as stale. Feel free to reopen if you revisit it.

PR #65, opened 2021-06-11
Change presence value of Optional

### LINKED ISSUE #64: IOptional.HasValue has confusing name.
**Description**

The naming of `IOptional.HasValue` implies that it is a value check, i.e. a nullability check.

**Steps to Reproduce**

```cs
var foo = new Optional<string?>(null);
if (foo.HasValue)
    Console.WriteLine(foo.Value.Length); // NullReferenceException
```

**Library / Runtime Information**

Remora.Discord.Core v2.2.1

**Resolution**

```cs
public interface IOptional
{

    [Obsolete("Prefer IOptional.IsSpecified")]
    public bool HasValue { get; }

    public bool IsSpecified { get; }
}

public readonly struct Optional<TValue> : IOptional
{

    [Obsolete("Prefer IOptional.IsSpecified")]
    public bool HasValue => IsSpecified; // For backwards compatibility

    public bool IsSpecified { get; }
}
```

Credit to @JakenVeina for pointing out this issue.

Foxtrek64:
Fixes #64

### DIFF
--- a/Backend/Remora.Discord.API/API/CDN.cs
+++ b/Backend/Remora.Discord.API/API/CDN.cs
@@ -643,7 +643,7 @@ namespace Remora.Discord.API
             Optional<ushort> imageSize = default
         )
         {
-            if (!application.CoverImage.IsSpecified || application.CoverImage.Value is null)
+            if (!application.CoverImage.IsSpecified)
             {
                 return new ImageNotFoundError();
             }
--- a/Backend/Remora.Discord.API/API/CDN.cs
+++ b/Backend/Remora.Discord.API/API/CDN.cs
@@ -643,7 +643,7 @@ namespace Remora.Discord.API
             Optional<ushort> imageSize = default
         )
         {
-            if (!application.CoverImage.HasValue)
+            if (!application.CoverImage.IsSpecified || application.CoverImage.Value is null)
             {
                 return new ImageNotFoundError();
             }
--- a/Backend/Remora.Discord.API/API/CDN.cs
+++ b/Backend/Remora.Discord.API/API/CDN.cs
@@ -643,7 +643,7 @@ namespace Remora.Discord.API
             Optional<ushort> imageSize = default
         )
         {
-            if (!application.CoverImage.IsSpecified)
+            if (!application.CoverImage.HasValue)
             {
                 return new ImageNotFoundError();
             }
--- a/Backend/Remora.Discord.Core/IOptional.cs
+++ b/Backend/Remora.Discord.Core/IOptional.cs
@@ -41,5 +41,10 @@ namespace Remora.Discord.Core
         /// Gets a value indicating whether the optional value is logically present.
         /// </summary>
         bool IsSpecified { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the value is <c>null</c>. If <see cref="IsSpecified"/> is false, <see cref="IsNull"/> is undefined.
+        /// </summary>
+        bool IsNull { get; }
     }
 }
--- a/Backend/Remora.Discord.Core/IOptional.cs
+++ b/Backend/Remora.Discord.Core/IOptional.cs
@@ -20,6 +20,7 @@
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 
+using System;
 using JetBrains.Annotations;
 
 namespace Remora.Discord.Core
@@ -33,6 +34,12 @@ namespace Remora.Discord.Core
         /// <summary>
         /// Gets a value indicating whether the optional contains a value.
         /// </summary>
+        [Obsolete("Prefer IOptional.IsSpecified")]
         bool HasValue { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the optional value is logically present.
+        /// </summary>
+        bool IsSpecified { get; }
     }
 }
--- a/Backend/Remora.Discord.Core/Optional.cs
+++ b/Backend/Remora.Discord.Core/Optional.cs
@@ -40,14 +40,14 @@ namespace Remora.Discord.Core
     [PublicAPI, DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
     public readonly struct Optional<TValue> : IOptional
     {
-        private readonly TValue? _value;
+        private readonly TValue _value;
 
         /// <summary>
         /// Gets the value contained in the optional.
         /// </summary>
         /// <exception cref="InvalidOperationException">Thrown if the optional does not contain a value.</exception>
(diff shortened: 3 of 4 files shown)


Backend/Remora.Discord.Core/Optional.cs
```diff
@@ -60,20 +60,18 @@ namespace Remora.Discord.Core
             }
         }
 
-        // Technically, this is a lie, but since the nullability of the type is reliant on the actual generic type and
-        // its annotations, this warning can be disabled.
-        #pragma warning disable CS8775
```
Foxtrek64, line 65: No warning needs to be disabled here. Works fine on NetFx and NetCore with it on. Comment is also removed since this the associated value has nothing to do with the nullability of the type, only whether it is logically present.

Backend/Remora.Discord.Core/Optional.cs
```diff
@@ -60,20 +60,18 @@ namespace Remora.Discord.Core
...
         /// <inheritdoc />
-        [MemberNotNullWhen(true, nameof(Value))]
```
Foxtrek64, line 67: Value is never null. It is initialized to `false` and only set high when a value is provided in the relevant constructor.

Foxtrek64:
Build failures seem to be unrelated.

```
/home/travis/build/Nihlus/Remora.Discord/Backend/Remora.Discord.API/API/CDN.cs(651,59): error CS8604: Possible null reference argument for parameter 'coverHash' in 'Result<Uri> CDN.GetApplicationCoverUrl(Snowflake applicationID, IImageHash coverHash, Optional<CDNImageFormat> imageFormat = default(Optional<CDNImageFormat>), Optional<ushort> imageSize = default(Optional<ushort>))'. [/home/travis/build/Nihlus/Remora.Discord/Backend/Remora.Discord.API/Remora.Discord.API.csproj]
```

Backend/Remora.Discord.API/API/CDN.cs
```diff
@@ -643,7 +643,7 @@ public static Result<Uri> GetApplicationCoverUrl
             Optional<ushort> imageSize = default
         )
         {
-            if (!application.CoverImage.HasValue)
+            if (!application.CoverImage.IsSpecified || application.CoverImage.Value is null)
```
JakenVeina, line 646: And here we have a great example of the other half of this problem: `.CoverImage` here is defined as `Optional<IImageHash>`. In reality, there is no possibility for this to be `null`, and `Optional<T>` shouldn't introduce one. I.E. `Optional<T>.Value` should be `T` not `T?`.

Foxtrek64: I've restored some things temporarily-

It seems `Optional.HasValue` is used in place of nullability checks in cases just like this all around the code.

Backend/Remora.Discord.Core/IOptional.cs
```diff
@@ -33,6 +34,17 @@ public interface IOptional
...
+
+        /// <summary>
+        /// Gets a value indicating whether the optional value is logically present.
+        /// </summary>
+        bool IsSpecified { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the value is <c>null</c>. If <see cref="IsSpecified"/> is false, <see cref="IsNull"/> is undefined.
+        /// </summary>
+        bool IsNull { get; }
```
Nihlus, line 48: An `Optional<T>` is not concerned with whether its contained value is null (or, in fact, if it is nullable at all). It is up to the declaring and using individual to appropriately check things based on the generic type. I don't see a compelling reason to add this property.

Backend/Remora.Discord.Core/IOptional.cs
```diff
@@ -33,6 +34,17 @@ public interface IOptional
         /// <summary>
         /// Gets a value indicating whether the optional contains a value.
         /// </summary>
+        [Obsolete("Prefer IOptional.IsSpecified")]
         bool HasValue { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the optional value is logically present.
+        /// </summary>
+        bool IsSpecified { get; }
```
Nihlus, line 43: `Optional<T>.HasValue` is named that way for two reasons - one, it *is* checking for the presence of a value: any value, including `null`. `null` is typically used to denote the absence of a value, yes, but it is very much a distinct value in and of itself. Two, it's named the same as `Nullable<T>.HasValue` for consistency and to follow the principle of least surprise.

`Optional<T>` mirrors `Nullable<T>` in a lot of ways, and I see no compelling reason to rename this property.

Backend/Remora.Discord.Core/Optional.cs
```diff
@@ -60,25 +60,34 @@ namespace Remora.Discord.Core
             }
         }
 
+        // CS8775: Member must have a non-null value when exiting in some condition.
         // Technically, this is a lie, but since the nullability of the type is reliant on the actual generic type and
         // its annotations, this warning can be disabled.
         #pragma warning disable CS8775
         /// <inheritdoc />
+        [Obsolete("Prefer Optional{TValue}.IsSpecified")]
         [MemberNotNullWhen(true, nameof(Value))]
```
Nihlus, line 69: This attribute, and the corresponding warning disabling above, does have room for improvement. I realize this isn't an optimal way of doing things, but it produces correct results in the majority of cases (e.g, when T is not a T?, be it struct or class). It is an attempt to convince the nullability annotations to warn correctly for the most common usages; I fully admit it's not always correct, though.

My preference would be to remove this attribute and figure out a better way of producing the correct nullability warnings depending on the contained type.

Nihlus (changes requested):

Nihlus:
I played around with just removing the annotations completely in Nihlus/Remora.Discord/better-optional - please have a peep and see if that might not be a better way to go with this.

This change only leaves nullable structs uncaught by nullability annotations, and that's not something we can really do much about (unfortunately).

JakenVeina:
> This change only leaves nullable structs uncaught by nullability annotations, and that's not something we can really do much about (unfortunately).

I'm not sure what you mean, but like you said above, an `Optional<T>` is not concerned with whether its contained value is null (or, in fact, if it is nullable at all). If `T` is nullable, it's up to the consumer to do appropriate null checks, after doing appropriate presence checks.

Backend/Remora.Discord.Core/IOptional.cs
JakenVeina: This might be a useful extension property to write for `Optional<T?> where T : struct` and `Optional<T?> where T : class` types, but even then, it probably doesn't belong in Remora itself.

Nihlus:
I'll consider this issue closed with the release at c39b05d, where I merged the mentioned improvements to `Optional<T>`'s nullability annotations.

PR #71, opened 2021-07-05
Enhanced bulk overwriting of slash command definitions

JakenVeina:
Enhanced the Rest Applications API to include the "default_permission" parameter when bulk overwriting commands, and to reduce verbosity.

### DIFF
--- a/Backend/Remora.Discord.API.Abstractions/API/Objects/Interactions/IApplicationCommandOverwriteData.cs
+++ b/Backend/Remora.Discord.API.Abstractions/API/Objects/Interactions/IApplicationCommandOverwriteData.cs
@@ -1,0 +1,49 @@
+﻿//
+//  IApplicationCommandOverwriteData.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) 2017 Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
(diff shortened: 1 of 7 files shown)


Backend/Remora.Discord.API.Abstractions/API/Objects/Interactions/IApplicationCommandOverwriteData.cs
```diff
@@ -0,0 +1,49 @@
...
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+using System.Collections.Generic;
+
```
Nihlus, line 24: You've got some spurious newlines here.

Backend/Remora.Discord.API.Abstractions/API/Objects/Interactions/IApplicationCommandOverwriteData.cs
```diff
@@ -0,0 +1,49 @@
...
+using Remora.Discord.Core;
+
+namespace Remora.Discord.API.Abstractions.Objects
+{
+    /// <summary>
+    /// Represents a request to create or update an application command.
+    /// </summary>
+    [PublicAPI]
+    public interface IApplicationCommandOverwriteData
```
Nihlus, line 35: I'm kind of on the fence about adding this interface. It implies that Discord has defined a matching datatype, but they haven't - they just refer to the existing command structure. On the other hand, there's some minor precedent for adding interfaces that represent a subset of Discord's data... I dunno, I'll have to sleep on this.

JakenVeina: Yeah, I had that same thought, and I ultimately went with adding it because A) it's not the only model type that doesn't DIRECTLY correspond to a Discord data model, and B) you seem to really really like interfaces.

Backend/Remora.Discord.API.Abstractions/API/Objects/Interactions/IApplicationCommandOverwriteData.cs
JakenVeina: My bad. Force of habit.

PR #75, opened 2021-07-10
Fixed incorrect optional/nullable modeling for Webhook.ApplicationID

JakenVeina:


Cisien:
Can confirm this works when I bypass the API wrapper and make the http call directly with the DiscordHttpClient

Tests/Remora.Discord.Tests/Samples/Objects/WEBHOOK/2021-07-10.json
```diff
@@ -0,0 +1,17 @@
+{
```
Nihlus, line 1: Run this file through the scrubber first.

Nihlus (changes requested):

Tests/Remora.Discord.Tests/Samples/Objects/WEBHOOK/2021-07-10.json
JakenVeina: Ah, right. I forgot about that. I asked Cisien first if he was cool with the sample being saved, but that makes more sense.

JakenVeina: Force-pushed to eliminate the unscrubbed file from history.

PR #76, opened 2021-07-14
Fix service lifetimes

Nihlus:
This PR corrects a number of mistakes in DI service registrations related to lifetimes. In a nutshell, a lot of things are no longer singletons, and HttpClient is being used in a more correct manner.

This fixes the issues we've been having with request customizations, as well as some REST-related problems that might not have been evident.

### DIFF
--- a/Backend/Remora.Discord.Rest/API/Applications/DiscordRestApplicationAPI.cs
+++ b/Backend/Remora.Discord.Rest/API/Applications/DiscordRestApplicationAPI.cs
@@ -20,6 +20,7 @@
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json;
@@ -55,6 +56,12 @@ namespace Remora.Discord.Rest.API
         {
             _discordHttpClient = discordHttpClient;
             _jsonOptions = jsonOptions.Value;
+        }
+
+        /// <inheritdoc cref="DiscordHttpClient.WithCustomization"/>
+        public DiscordRequestCustomization WithCustomization(Action<RestRequestBuilder> requestCustomizer)
+        {
+            return _discordHttpClient.WithCustomization(requestCustomizer);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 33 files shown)


Backend/Remora.Discord.Rest/API/Applications/DiscordRestApplicationAPI.cs
```diff
@@ -57,6 +58,12 @@ IOptions<JsonSerializerOptions> jsonOptions
             _jsonOptions = jsonOptions.Value;
         }
 
+        /// <inheritdoc cref="DiscordHttpClient.WithCustomization"/>
```
JakenVeina, line 61: Don't think it'd be a appropriate to put all these on a base class?

Backend/Remora.Discord.Rest/DiscordHttpClient.cs
```diff
@@ -68,7 +68,7 @@ IOptions<JsonSerializerOptions> serializerOptions
         /// </summary>
         /// <param name="requestCustomizer">The action that customizes the request.</param>
         /// <returns>The created customization.</returns>
-        public DiscordRequestCustomization AddCustomization(Action<RestRequestBuilder> requestCustomizer)
+        public DiscordRequestCustomization WithCustomization(Action<RestRequestBuilder> requestCustomizer)
```
JakenVeina, line 71: So, wait, after an hour's worth of arguing, this ALREADY EXISTS?!

Why the name change here?

Backend/Remora.Discord.Rest/API/Applications/DiscordRestApplicationAPI.cs
Nihlus: That's probably a good idea, yeah.

Backend/Remora.Discord.Rest/DiscordHttpClient.cs
Nihlus: I thought I made that clear, and that it was because of its existence that I originally wanted to expose `DiscordHttpClient` as the main modification point.

The name change is just personal preference. I think `WithCustomization` works better than `AddCustomization`, but it's either or.

JakenVeina: I dunno, that must've been before I joined in the convo.

I also like that name better, and I just realized it's not a breaking change, cause that addition hasn't been published yet, so nevermind.

PR #77, opened 2021-07-15
Enhanced Optional<T> to support meaningful stringification.

JakenVeina:
So, hear me out. I encountered the need to write the following while building out logging infrastructure...

```cs
=> _userTracking.Invoke(
    logger,
    userId,
    guildId,
    username.HasValue
        ? username.Value.ToString()
        : "Unspecified",
    discriminator.HasValue
        ? discrimincator.Value.ToString()
        : "Unspecified",
    avatarHash.HasValue
        ? avatarHash.Value.ToString()
        : "Unspecified",
    nickname.HasValue
        ? nickname.Value.ToString()
        : "Unspecified");
```

This is necessary because I can't reasonably configure all logger sinks to know how to serialize an `Optional<T>` value, especially those that attempt to JSON-serialize objects. Instead, they'll try and read from `.Value` and cause an exception in logging.

I'm usually a fan of not having debugging logic embedded in a type, so I generally like `DebuggerDisplay` myself, but for a type that is borderline "primitive" I think there's a fair bit of value in giving it a somewhat-meaningful `.ToString()` implementation.

### DIFF
--- a/Backend/Remora.Discord.Core/Optional.cs
+++ b/Backend/Remora.Discord.Core/Optional.cs
@@ -23,7 +23,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using JetBrains.Annotations;
 
 namespace Remora.Discord.Core
@@ -37,7 +36,7 @@ namespace Remora.Discord.Core
     /// value would never be serialized, but a nullable with a null value would (albeit as "null").
     /// </summary>
     /// <typeparam name="TValue">The inner type.</typeparam>
-    [PublicAPI, DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
+    [PublicAPI]
     public readonly struct Optional<TValue> : IOptional
     {
         private readonly TValue _value;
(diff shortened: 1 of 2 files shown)


Backend/Remora.Discord.Core/Optional.cs
```diff
@@ -136,5 +121,13 @@ public override int GetHashCode()
         {
             return HashCode.Combine(_value, this.HasValue);
         }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            return this.HasValue
+                ? $"{{{_value?.ToString() ?? "null"}}}"
```
Nihlus, line 129: What's with the string interpolation here?

JakenVeina: Uhh... it's a way to build strings?

Nihlus: Yes, but why :P What's the point of adding the curly braces around the value?

JakenVeina: It provides a lot of subtle visual clues about the fact that this isn't just a "pure" value, but it's wrapped in a container. Pretty darned minor, but I like the idea. For example, it provides a distinction between an empty optional value, and an optional value containing the word "Empty". Also, from a live use-case...

Capture


Here, it also provides a subtle indication that some of these logged values COULD have been unspecified, but weren't.

Nihlus: Fair enough!

PR #80, opened 2021-07-28
Updated the RequireContextAttribute and RequireContextCondition to better fine tune the command accessibility context.

ProIcons:


Remora.Discord.Commands/Conditions/Attributes/RequireContextAttribute.cs
```diff
@@ -32,17 +33,33 @@ namespace Remora.Discord.Commands.Conditions
...
-        /// Gets the command context.
+        /// Gets the command grouped contexts.
         /// </summary>
-        public ChannelContext Context { get; }
+        public ChannelContext[]? ChannelContexts { get; }
+
+        /// <summary>
+        /// Gets the command individual channel type contexts.
+        /// </summary>
+        public ChannelType[]? ChannelTypes { get; }
```
Nihlus, line 43: You can remove one of these arrays and make it non-nullable by converting the constructor parameters to the appropriate types. Personally, I think it'd be most appropriate to remove `ChannelContexts` and convert each grouped context into its constituent parts. Saves on logic and lets users have a simpler API, while at the same time allowing greater detailed control :)

Remora.Discord.Commands/Conditions/Attributes/RequireContextAttribute.cs
```diff
@@ -32,17 +33,33 @@ namespace Remora.Discord.Commands.Conditions
...
     {
         /// <summary>
-        /// Gets the command context.
+        /// Gets the command grouped contexts.
         /// </summary>
-        public ChannelContext Context { get; }
+        public ChannelContext[]? ChannelContexts { get; }
+
+        /// <summary>
+        /// Gets the command individual channel type contexts.
```
Nihlus, line 41: Improve this documentation comment. It doesn't make much sense.

Remora.Discord.Commands/Conditions/RequireContextCondition.cs
```diff
@@ -68,21 +69,42 @@ public async ValueTask<Result> CheckAsync(RequireContextAttribute attribute, Can
 
             var channel = getChannel.Entity;
 
-            return attribute.Context switch
+            if (attribute.ChannelTypes != null)
```
Nihlus, line 72: Use `is null` and `is not null` where appropriate, depending on which property you keep.

Remora.Discord.Commands/Conditions/RequireContextCondition.cs
```diff
@@ -68,21 +69,42 @@ public async ValueTask<Result> CheckAsync(RequireContextAttribute attribute, Can
...
+            {
+                var conditionResult = channelContext switch
+                {
+                    ChannelContext.DM => channel.Type is DM
+                        ? Result.FromSuccess()
+                        : new ConditionNotSatisfiedError("This command can only be used in a DM."),
+                    ChannelContext.GroupDM => channel.Type is GroupDM
+                        ? Result.FromSuccess()
+                        : new ConditionNotSatisfiedError("This command can only be used in a group DM."),
+                    ChannelContext.Guild => channel.Type is GuildText or GuildVoice or GuildCategory or GuildNews
```
Nihlus, line 94: There's probably an argument to be made for moving these groupings out into an extension method on `ChannelContext` or a dictionary of some kind.

Nihlus (changes requested):

Remora.Discord.Commands/Conditions/Attributes/RequireContextAttribute.cs
ProIcons: Removing the ChannelContexts will be a breaking change, however i have no issue on doing that but i didn't understand that part:
> I think it'd be most appropriate to remove ChannelContexts and **convert each grouped context into its constituent parts**.

ProIcons:
I updated what you asked however if my descriptions/comments aren't that good, I'm really sorry I'm trying :)

PR #86, opened 2021-08-01
[WIP] Rework pre- and post-execution event API

Nihlus:
This PR reworks the pre- and post-execution event API, making it a bit less tedious to work with. Furthermore, it also alters `InteractionContext`, providing more data useful to both command implementations and execution events.


ProIcons:
Looks good to me, makes sense.

PR #87, opened 2021-08-06
Format Helper Methods

Hamsterland:
# Description
Adds helper methods to format text.

## Supported Formats
- [x] Basic Markdown
- [x] Markdown Timestamps
- [x] Mentions (users, channels, roles)

## Tests
- [x] Markdown Tests
- [x] Mention Tests


Hamsterland:
Any other formatting considerations?

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
```diff
@@ -0,0 +1,77 @@
...
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+using System;
+
+namespace Remora.Discord.API.Format
```
Nihlus, line 25: I would probably name this `Formatting` instead of `Format`, since that's what we're implementing.

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// </summary>
+        /// <param name="dateTime">The time to format.</param>
+        /// <param name="timestampStyle">The style to format into.</param>
+        /// <returns>
+        /// A Discord markdown-formatted Timestamp string.
+        /// </returns>
+        public static string Timestamp(DateTime dateTime, TimestampStyle timestampStyle = TimestampStyle.Default)
+            => Timestamp(((DateTimeOffset)dateTime).ToUnixTimeSeconds(), timestampStyle);
+
+        private static char TimestampStyleToCode(TimestampStyle timestampStyle)
```
Nihlus, line 62: Document private methods as well.

Backend/Remora.Discord.API/API/Format/Markdown.cs
```diff
@@ -0,0 +1,119 @@
...
+
+using JetBrains.Annotations;
+
+namespace Remora.Discord.API.Format
+{
+    /// <summary>
+    /// Provides helper methods to format strings into Discord Markdown.
+    /// </summary>
+    [PublicAPI]
+    public static partial class Markdown
```
Nihlus, line 31: Are there any escape sequences we need to take into account for these?

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
```diff
@@ -0,0 +1,77 @@
...
+            return timestampStyle switch
+            {
+                TimestampStyle.ShortTime => 'f',
+                TimestampStyle.LongTime => 'F',
+                TimestampStyle.ShortDate => 'd',
+                TimestampStyle.LongDate => 'D',
+                TimestampStyle.ShortDateTime => 'f',
+                TimestampStyle.LongDateTime => 'F',
+                TimestampStyle.RelativeTime => 'R',
+                _ => TimestampStyleToCode(TimestampStyle.Default)
```
Nihlus, line 73: It may be prudent to add a comment here to clarify that this won't cause infinite recursion.

Backend/Remora.Discord.API/API/Format/TimestampStyle.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// 2 months ago.
+        /// </summary>
+        RelativeTime,
+
+        /// <summary>
+        /// The default timestamp style.
+        /// </summary>
+        Default = ShortDateTime
```
Nihlus, line 68: Since Discord allows you to omit the format specifier altogether, perhaps it would be better to remove this and make the style arguments nullable (with `= default`)?

Nihlus (changes requested):
Nice work! This will be a very useful addition.

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// A Discord markdown-formatted Timestamp string.
+        /// </returns>
+        public static string Timestamp(DateTime dateTime, TimestampStyle timestampStyle = TimestampStyle.Default)
+            => Timestamp(((DateTimeOffset)dateTime).ToUnixTimeSeconds(), timestampStyle);
+
+        private static char TimestampStyleToCode(TimestampStyle timestampStyle)
+        {
+            return timestampStyle switch
+            {
+                TimestampStyle.ShortTime => 'f',
```
carlst99, line 66:
```suggestion
                TimestampStyle.ShortTime => 't',
```
ShortTime and LongTime are 't' and 'T' respectively 🙂

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
```diff
@@ -0,0 +1,77 @@
...
+                TimestampStyle.LongTime => 'F',
```
carlst99, line 67:
```suggestion
                TimestampStyle.LongTime => 'T',
```

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
Hamsterland: Resolved. Thanks

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
Hamsterland: Thanks for pointing this out!

Backend/Remora.Discord.API/API/Format/Markdown.Timestamps.cs
Hamsterland: Comment added.

Backend/Remora.Discord.API/API/Format/TimestampStyle.cs
Hamsterland: Done!

Backend/Remora.Discord.API/API/Format/Markdown.cs
Hamsterland: Is it the library's responsibility to take care of this? Personally, I don't think so. The API consumer should just provide the input and the library should add Discord's markdown around it.

Nihlus: I'm kind of split on this as well. We can leave it for later :)

Backend/Remora.Discord.API/API/Formatting/Markdown.Timestamps.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                TimestampStyle.ShortTime => 't',
+                TimestampStyle.LongTime => 'T',
+                TimestampStyle.ShortDate => 'd',
+                TimestampStyle.LongDate => 'D',
+                TimestampStyle.ShortDateTime => 'f',
+                TimestampStyle.LongDateTime => 'F',
+                TimestampStyle.RelativeTime => 'R',
+
+                // This will not cause infinite recursion.
```
Nihlus, line 85: There's no need for this comment now.

Nihlus (changes requested):

Nihlus:
I'm happy with the implementations now - great work! As soon as you implement tests for them, I'm ready to merge.

Backend/Remora.Discord.API/API/Formatting/Markdown.Timestamps.cs
Hamsterland: Removed, thanks!

Backend/Remora.Discord.API/API/Formatting/Markdown.cs
```diff
@@ -0,0 +1,119 @@
...
+
+using JetBrains.Annotations;
+
+namespace Remora.Discord.API.Formatting
+{
+    /// <summary>
+    /// Provides helper methods to format strings into Discord Markdown.
+    /// </summary>
+    [PublicAPI]
+    public static partial class Markdown
```
Tylertron1998, line 31: How about if the content already contains certain markdown characters - perhaps escaping those is a good idea

Backend/Remora.Discord.API/API/Formatting/Mention.cs
```diff
@@ -0,0 +1,61 @@
...
+    public static class Mention
+    {
+        /// <summary>
+        /// Creates a mention string for a user.
+        /// </summary>
+        /// <param name="snowflake">The user Snowflake ID.</param>
+        /// <returns>
+        /// A user mention string.
+        /// </returns>
+        public static string User(Snowflake snowflake) => $"<@{snowflake.Value}>";
```
Tylertron1998, line 41: There is also a member mention (it mentions the nickname specifically, where as user mentions the username even if they have a nickname) if I remember right, it's <@!id>

Nihlus: There is, but it doesn't do anything different than the other one. Discord devs themselves recommend just using the straight mention.

Hamsterland: This only applies to mobile push notifications anyway afaik, not really worth implementing

Backend/Remora.Discord.API/API/Formatting/Markdown.cs
```diff
@@ -0,0 +1,119 @@
...
+        public static string InlineCode(string text) => $"`{text}`";
+
+        /// <summary>
+        /// Formats a string to use Markdown Block Code formatting.
+        /// </summary>
+        /// <param name="text">The input text to format.</param>
+        /// <returns>
+        /// A markdown-formatted block code string.
+        /// </returns>
+        public static string BlockCode(string text) => $"```\n{text}\n```";
```
AraHaan, line 94: I feel like this should call into ``BlockCode(test, string.Empty)`` instead.

Backend/Remora.Discord.API/API/Formatting/Markdown.cs
```diff
@@ -0,0 +1,119 @@
...
+        public static string BlockCode(string text, string language) => $"```{language}\n{text}\n```";
+
+        /// <summary>
+        /// Formats a string to use Markdown Quote Block formatting.
+        /// </summary>
+        /// <param name="text">The input text to format.</param>
+        /// <returns>
+        /// A markdown-formatted block quote string.
+        /// </returns>
+        public static string BlockQuote(string text) => $">>> {text}";
```
AraHaan, line 117: Can an ``BlockList`` method be added as well?

It could possibly be something like so:

```cs
public static string BlockList(string text, string mark = "-") => $" {mark} {text}\n";
```
Where mark can be: ``-``, a number (for number based lists that are supported in markdown), an ``*`` for lists with the dots, etc.

Also I think this line in ``BlockQuote`` can be:

```suggestion
        public static string BlockQuote(string text) => $">>> {text}\n";
```

AraHaan (changes requested):

Backend/Remora.Discord.API/API/Formatting/Markdown.cs
Nihlus: Discord doesn't recognize lists in its markdown, so that would be out of scope.

Rutherther:
What about adding formatters for `IEmote` and `IPartialEmote` to format that can be sent to Discord?

That would mean for guild emojis (those that have set ID), it would be `<:id:name>`, for others just `name` that corresponds to unicode emoji. 

Sorry if this is somewhere in Remora already, I tried asking on Discord, but no one replied that it is, so I assume it is not.

Nihlus:
> What about adding formatters for `IEmote` and `IPartialEmote` to format that can be sent to Discord?
> 
> That would mean for guild emojis (those that have set ID), it would be `<:id:name>`, for others just `name` that corresponds to unicode emoji.
> 
> Sorry if this is somewhere in Remora already, I tried asking on Discord, but no one replied that it is, so I assume it is not.

This would be a good addition to this, yeah :)

Hamsterland:
I'm going to remove the Draft status from this PR. 

- jer#9999 has offered to write tests
- someone else can PR the emote parses in later, I can't work on that at the moment

Hamsterland:
> What about adding formatters for `IEmote` and `IPartialEmote` to format that can be sent to Discord?
> 
> That would mean for guild emojis (those that have set ID), it would be `<:id:name>`, for others just `name` that corresponds to unicode emoji.
> 
> Sorry if this is somewhere in Remora already, I tried asking on Discord, but no one replied that it is, so I assume it is not.
Would it look something like this?
```cs
I understand this working for an `IEmoji`, but if the optionals are not guaranteed to be specified for an `IPartialEmoji`, what should occur? An exception?

**Comment by Hamsterland** [comment]
baa996b4a3549d397f92b5c42b02dfad797eecf0 Added support for Username/Nickname mentioning.

**Comment by Rutherther** [comment]
> > What about adding formatters for `IEmote` and `IPartialEmote` to format that can be sent to Discord?
> > That would mean for guild emojis (those that have set ID), it would be `<:id:name>`, for others just `name` that corresponds to unicode emoji.
> > Sorry if this is somewhere in Remora already, I tried asking on Discord, but no one replied that it is, so I assume it is not.
> 
> I understand this working for an `IEmoji`. But if the optionals are not guaranteed to be specified for an `IPartialEmoji`, what should occur? An exception? A result?

I think Result will be appropriate, yeah

**Comment by Nihlus** [comment]
As far as I know, Discord doesn't render `<@...>` vs `<@!...>` any differently anymore - it's a leftover from a previous client version. They recommendonly using `<@...>` for new code.

**Comment by Nihlus** [comment]
I'd prefer not to add functionality that's marked as obsolete from the get-go - obsolete stuff just plain shouldn't be used.

**Comment by Nihlus** [comment]
Still waiting on tests for this PR, but I'd love to merge it!

**Comment by Hamsterland** [comment]
> Still waiting on tests for this PR, but I'd love to merge it!

In which folder would tests come under?

**Comment by Nihlus** [comment]
https://github.com/Nihlus/Remora.Discord/tree/master/Tests/Remora.Discord.API.Tests, and then some appropriately named folder.

**Comment by Foxtrek64** [comment]
@Nihlus @Hamsterland Should we move these helpers into the new `Remora.Discord.Extensions` library?

**Comment by Foxtrek64** [comment]
Also, a note on a few things to add:

1. Url escapes to prevent embedding.
```cs
public static string EscapeUrl(string url) => $"<{url}>";
```
2. Markdown hyperlinks for embeds. Don't know if Discord supports tooltips but if so that should be added.
```cs
public static string Hyperlink(string url) => Url(url, url);
public static string Hyperlink(string text, string url) => $"[{text}]({url})";
```
3. Sanitation utilities
```cs
private static readonly Regex UserMentionRegex = new("<@!?(?<Id>[0-9]+)>", RegexOptions.Compiled);
private static readonly Regex RoleMentionRegex = new("<@&(?<Id>[0-9]+)>", RegexOptions.Compiled);
private static readonly string[] SensitiveCharacters = { "\\", "*", "_", "~", "`", "|", ">" };

public static string SanitizeEveryone(string text)
    => text.Replace("@everyone", "@\x200beveryone")
            .Replace("@here", "@\x200bhere");

public static string SanitizeUserMentions(string text)
    => UserMentionRegex.Replace(text, "<@\x200b${Id}>");

public static string SanitizeRoleMentions(string text)
    => RoleMentionRegex.Replace(text, "<@&\x200b${Id}>");

public static string Sanitize(string text)
    => SensitiveCharacters.Aggregate(text, (current, unsafeChar) => current.Replace(unsafeChar, $@"\{unsafeChar}"));

**Review by Foxtrek64 — APPROVED** [review]
Lgtm

**Comment by Hamsterland** [comment]
Just waiting on tests

> https://github.com/Nihlus/Remora.Discord/tree/master/Tests/Remora.Discord.API.Tests, and then some appropriately named folder.

Do we want an Extensions test project now too?

**Comment by Nihlus** [comment]
Here's something that would be great to add to this PR - ANSI colour codes are now supported in Discord!
https://gist.github.com/kkrypt0nn/a02506f3712ff2d1c8ca7c9e0aed7c06

**Comment by Hamsterland** [comment]
I apologise for the delay, I am continuing work on this now

File: Remora.Discord.Extensions/Formatting/Markdown.Timestamps.cs [hunk 13]
```diff
@@ -0,0 +1,89 @@
...
+    {
+        /// <summary>
+        /// Formats a Unix timestamp value into Discord Markdown Timestamp.
+        /// </summary>
+        /// <param name="unixTimestamp">The Unix timestamp to format.</param>
+        /// <param name="timestampStyle">The style to format into.</param>
+        /// <returns>
+        /// A Discord markdown-formatted Timestamp string.
+        /// </returns>
+        public static string Timestamp(long unixTimestamp, TimestampStyle? timestampStyle = default)
```
VelvetThePanda, line 37: Instead of making this nullable, this should default to short date and time
(:f) as it is the default in the client.

Remora.Discord.Extensions/Formatting/Markdown.Timestamps.cs
```diff
@@ -0,0 +1,89 @@
...
+            => Timestamp(((DateTimeOffset)dateTime).ToUnixTimeSeconds(), timestampStyle);
+
+        /// <summary>
+        /// Converts a <see cref="TimestampStyle"/> to its Discord Markdown code.
+        /// </summary>
+        /// <param name="timestampStyle">The style to convert.</param>
+        /// <returns>
+        /// The Discord Markdown code for the given style.
+        /// </returns>
+        private static char TimestampStyleToCode(TimestampStyle? timestampStyle)
```
VelvetThePanda, line 73: This exists, but the null check is never used in the above.

Remora.Discord.Extensions/Formatting/Markdown.cs
```diff
@@ -0,0 +1,200 @@
...
+        /// Escapes a URL to prevent embedding of its content.
+        /// </summary>
+        /// <param name="url">The URL to escape.</param>
+        /// <returns>
+        /// A formatted URL string that will not embed its content.
+        /// </returns>
+        public static string EscapeUrl(string url) => $"<{url}>";
+
+        /// <summary>
+        /// Formats a string to contain a Markdown hyperlink of itself. The hyperlink will only hyperlink inside
```
VelvetThePanda, line 146: Webhooks (e.g interactions) also have support for hyperlinks.

VelvetThePanda (changes requested):

Remora.Discord.Extensions/Formatting/Markdown.cs
Hamsterland: What do you mean? How would that affect this use case?

Remora.Discord.Extensions/Formatting/Markdown.Timestamps.cs
Hamsterland: This was how it was originally, but Jax suggested that it be done this way instead.

Remora.Discord.Extensions/Formatting/Markdown.Timestamps.cs
Hamsterland: Same as above

Remora.Discord.Extensions/Formatting/Markdown.Timestamps.cs
Nihlus: My original reasoning was that since Discord allows you to omit the style specifier, having no specifier as our default would be more appropriate. This will - at least to my knowledge - shift the responsibility of deciding what the default is to Discord, leaving us more flexible if they decide to change their minds later.

MazeXP:
> Here's something that would be great to add to this PR - ANSI colour codes are now supported in Discord code blocks! https://gist.github.com/kkrypt0nn/a02506f3712ff2d1c8ca7c9e0aed7c06

This could be integrated for the building of an ANSI formatted string:
https://gist.github.com/MazeXP/617fcf7c8d632cfbf3a7376753df6307

You can just include and modify it to fit in.
I am also open for suggestions on this one.

AraHaan:
could these color codes be also be used for colorizing select menu texts?

MazeXP:
> could those color codes be also be used for colorizing select menu texts? (the text above the description)

Nope, the select menu does not support markdown syntax in its options.

Hamsterland:
@MazeXP that sounds like an excellent addition if you’re willing to add it to this library. I think you should open a separate PR for that if you do.

AraHaan:
@MazeXP I tried that but it did not work even when I copied it from raw.

MazeXP:
What does it look like on your end?

AraHaan:
image

LOL I just realized it, I copied the whole thing.

AraHaan:
Violot looks kind of blue to me.

MazeXP:
The color names are plainly based on the Solarized Palette that Discord is using for formatting as well.
Therefore the names could be freely changed afterwards.

MazeXP:
I will setup a draft PR with the ANSI formatting string building then we can discuss on the color names there.

PR #91, opened 2021-08-17
Update thread-related permissions.

carlst99:
See https://github.com/discord/discord-api-docs/pull/3672 . Note Discord advises this change will not be deployed until (tentatively) September 7th, but guilds will be allowed to make the change earlier if desired.

I've also removed the `Flags` attribute from the `DiscordPermission` enum. I believe this is correct; you can't perform bitwise operations on it and it is now consistent with the attribute markups of the other permission enums.


Nihlus:
Nice, thank you. I'll hold off on merging until discord/discord-api-docs#3672 is merged.

carlst99:
This change has been merged to the docs now 👍.

PR #98, opened 2021-08-24
Fix send/receive gateway tasks never completing

Cyberboss:
If RunAsync was called with an already cancelled token, this delay would throw an improperly handled TaskCanceledException. As a result, regular cleanup would be bypassed.

Also fixed certain transport service disconnect calls indicating they will reconnect when they won't.

This ensures the path of disconnection is followed as if the gateway itself requested it.

### DIFF
--- a/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
+++ b/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
@@ -231,7 +231,7 @@ namespace Remora.Discord.Gateway
 
                     if (_transportService.IsConnected)
                     {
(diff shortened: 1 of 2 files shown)


AraHaan:
Why not invoke ``RunAsync()`` with ``default``?

I can properly run my bot using ``default`` passed in like so ``RunAsync(default).ConfigureAwait(false);``.

Cyberboss:
My software needs to support many instances of bots and be able to online/offline them at will so using a `default` `CancellationToken` is out of the question.

Porting from Discord.Net, my integration tests picked this up as it's possible for said CancellationToken to already be cancelled once `RunAsync` is invoked.

Nihlus:
Good catches, thanks!

PR #100, opened 2021-08-29
Update tutorial guide

Nonparoxysmic:
This makes some improvements and fixes some minor issues in the Getting Started guide.


docs/custom/guides/getting-started.md
```diff
@@ -1,6 +1,4 @@
-Title: Getting Started
```
Nihlus, line 1: This is used by the Wyam docs generator as title metadata, and should not be altered.

Nihlus (changes requested):
Thanks! One minor change, then we're good to merge.

Nihlus (approved):

PR #116, opened 2021-10-02
Fix incorrect handling of target_application in invites API

### LINKED ISSUE #117: Creating a channel invite for an embedded application fails on JSON deserialization
**Description**

When you generate an invite for a channel with an embedded application included, the response's content will cause JSON deserialization to an ``IInvite`` to fail.

**Steps to Reproduce**

1. Using an implementation of ``IDiscordRestChannelAPI``, call ``CreateChannelInviteAsync`` including a target type of ``InviteTarget.EmbeddedApplication``, and include a target application ID.
2. The response will always be a failure with a JSON deserialization error. The error is on the ``TargetApplication`` field.

**Expected Behavior**

The deserialization should not fail.

**Current Behavior**

The deserialization fails.

**Library / Runtime Information**

``Remora.Discord 3.0.68``

bobbahbrown:
As noted in the Discord, the existing handling of invites from the API is invalid when it comes to ``target_application``. We have incorrectly typed this as a Snowflake, when in reality it is a partial application object.

You can find this definition on the Discord documentation, [here](https://discord.com/developers/docs/resources/invite#invite-object-invite-structure).

I've also gone ahead and added support for this field to the [Invite Create gateway event](https://discord.com/developers/docs/topics/gateway#invite-create), and corrected the (presumably changed?) field name, ``target_type``. Thanks to Falcon in the Discord for pointing this out.

Fixes #117.

### DIFF
--- a/Backend/Remora.Discord.API.Abstractions/API/Gateway/Events/Invites/IInviteCreate.cs
+++ b/Backend/Remora.Discord.API.Abstractions/API/Gateway/Events/Invites/IInviteCreate.cs
@@ -76,7 +76,12 @@ namespace Remora.Discord.API.Abstractions.Gateway.Events
         /// <summary>
         /// Gets the type of user target for this invite.
         /// </summary>
-        Optional<InviteTarget> TargetUserType { get; }
+        Optional<InviteTarget> TargetType { get; }
+
+        /// <summary>
+        /// Gets the embedded application this invite is for.
+        /// </summary>
+        Optional<IPartialApplication> TargetApplication { get; }
 
         /// <summary>
         /// Gets a value indicating whether the invite is temporary (invited users will be kicked on disconnect unless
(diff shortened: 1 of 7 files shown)


Backend/Remora.Discord.API.Abstractions/API/Gateway/Events/Invites/IInviteCreate.cs
```diff
@@ -76,7 +76,12 @@ public interface IInviteCreate : IGatewayEvent
         /// <summary>
         /// Gets the type of user target for this invite.
         /// </summary>
-        Optional<InviteTarget> TargetUserType { get; }
+        Optional<InviteTarget> TargetType { get; }
```
Nihlus, line 79: Try to keep these in the same order as they appear in the Discord docs (that is, target_type first and target_user after).

Nihlus (changes requested):

Backend/Remora.Discord.API.Abstractions/API/Gateway/Events/Invites/IInviteCreate.cs
bobbahbrown: Fixed, thanks

Nihlus (approved):

PR #127, opened 2021-11-26
Add DiscordGatewayClient#Latency

carlst99:
This introduces a `Latency` property to the `DiscordGatewayClient`. This property is updated upon receiving a heartbeat and calculated from the time taken between sending a heartbeat and it being acknowledged. Hence providing an estimate of round-trip latency to the gateway.


VelvetThePanda:
Much appreciated
party cat

PR #128, opened 2021-11-26
Omittable default responders

VelvetThePanda:
This PR adds an additional parameter to `AddDiscordCommands`, which simply skips over adding the default `AddCommandResponder` if set to false. 

In theory this should make it easier and potentially more efficient (perhaps a micro-optimization, but optimization is optimization) to stub in a responder of the end user's choosing, with the typical use-case being guild-specific prefixes.

Not sure if I should've written a test for this or anything, given it's something so simple, but do feel free to inform me if I've done something wrong.

thaaanks

### DIFF
--- a/Remora.Discord.Commands/Extensions/ServiceCollectionExtensions.cs
+++ b/Remora.Discord.Commands/Extensions/ServiceCollectionExtensions.cs
@@ -146,14 +146,14 @@ public static class ServiceCollectionExtensions
             return serviceCollection;
         }
 
-        serviceCollection.TryAddSingleton<SlashService>();
-        serviceCollection.AddInteractionResponder();
-        serviceCollection.AddAutocompleteProvider(typeof(EnumAutocompleteProvider<>));
-
         if (useDefaultInteractionResponder)
         {
-            serviceCollection.AddResponder<AutocompleteResponder>();
+            serviceCollection.AddInteractionResponder();
         }
+
+        serviceCollection.TryAddSingleton<SlashService>();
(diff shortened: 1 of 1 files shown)


Remora.Discord.Commands/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -112,7 +114,11 @@ public static IServiceCollection AddDiscordCommands
         );
 
         serviceCollection.AddCommands();
-        serviceCollection.AddCommandResponder();
+
```
Nihlus, line 117: This should also conditionally exclude `CommandInteractionResponder`, since both handle incoming commands. Optionally, add a second parameter for that responder as well.

Nihlus (changes requested):

Remora.Discord.Commands/Extensions/ServiceCollectionExtensions.cs
VelvetThePanda: Fair enough, thanks for the insight. ❤️

Nihlus: I think you moved the wrong responder :P AutocompleteResponder should always be registered - it is `AddInteractionResponder` you want to conditionally add.

Nihlus (changes requested):

VelvetThePanda:
Whoops. Should be fixed @Nihlus

PR #129, opened 2021-11-27
Add Voice Gateway Object Definitions

carlst99:
This PR introduces the types used by version four of the voice gateway. It includes the JSON converter (`VoicePayloadConverter`), and all relevant tests.

Some factors to note:

- The `VoiceGatewayCloseStatus.Ratelimited` enum value is undocumented by the API. I chose to include it regardless because ratelimiting needs to be considered, and this OP code should never change (or at least, be reused) for the lifetime of this version of the voice gateway.
- Two event definitions (`IVoiceSpeakingEvent` and `IVoiceClientDisconnect`) have been placed into the unstable library, due to them having an associated OP code but no documented type information.
- The `VoicePayloadConverter` returns a null value, rather than throwing an exception, when it encounters an unknown OP code during deserialization only. This is because the API sends OP codes that have been omitted from the current definitions due to a lack of documentation, and I'd rather the caller handle the null return appropriately than an exception being generated each time this happens.

I additionally changed the `ServiceCollectionExtensions.AddExperimentalDiscordApi(...)` method to accept a name for the `JsonSerializerOptions` object that it configures. This defaults to a value of "Discord" to maintain consistency with the primary configuration point, and ensures that objects in the unstable package can be deserialized without the user having to register them.

Finally, I think I've remembered to use the `var` keyword where appropriate :joy:. Famous last words...


Nihlus:
Adding this link to the WIP unified gateway as well, since this will hopefully be part of the eventual voice implementation: https://github.com/JakenVeina/DiscordGateway

Backend/Remora.Discord.API.Abstractions/API/VoiceGateway/Events/IVoiceReady.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Gets the synchronization source ID. This acts as an identifier for this voice connection.
+    /// </summary>
+    uint SSRC { get; }
+
+    /// <summary>
+    /// Gets the IP that the voice server is available on.
+    /// </summary>
+    string IP { get; }
```
Nihlus, line 42: This could probably be implemented with `IPAddress` from System.Net (https://docs.microsoft.com/en-us/dotnet/api/system.net.ipaddress?view=net-6.0)

Backend/Remora.Discord.Unstable/Unstable/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -40,16 +43,29 @@ public static class ServiceCollectionExtensions
...
         (
-            this IServiceCollection serviceCollection
+            this IServiceCollection serviceCollection,
+            string? optionsName = "Discord"
         )
         {
-            serviceCollection.Configure<JsonSerializerOptions>(jsonOptions =>
+            serviceCollection.Configure<JsonSerializerOptions>(optionsName, jsonOptions =>
             {
+                jsonOptions.Converters.Add(new UnstableVoicePayloadConverter());
```
Nihlus, line 56: You can use the extension method `AddConverter` for this - it'll insert it before all other converters in the list, ensuring it takes priority over existing ones.

Nihlus (changes requested):
This looks great! I'd like you to run the data scrubber against your new sample files before I merge, though, since some IDs can be considered sensitive data.

carlst99:
Thanks - scrubber has been run and review resolved.

Nihlus (approved):

PR #130, opened 2021-11-27
Add EmbedBuilder

Foxtrek64:
This PR adds a `Remora.Discord.Extensions` library and adds the first helper library, an `EmbedBuilder`.

* [x] Add EmbedBuilder
* [x] Add EmbedBuilder Tests

### DIFF
--- a/Remora.Discord.Extensions/Embeds/Constants.cs
+++ b/Remora.Discord.Extensions/Embeds/Constants.cs
@@ -30,34 +30,44 @@ namespace Remora.Discord.Extensions.Embeds
     public static class Constants
     {
         /// <summary>
-        /// Returns the maximum number of fields allowed by Discord.
-        /// </summary>
-        public const int MaxFieldCount = 25;
-
-        /// <summary>
-        /// Returns the maximum title length allowed by Discord.
+        /// Gets the maximum length of a title.
         /// </summary>
         public const int MaxTitleLength = 256;
 
         /// <summary>
-        /// Returns the maximum description length allowed by Discord.
+        /// Gets the maximum length of a description.
         /// </summary>
-        public const int MaxDescriptionLength = 2048;
+        public const int MaxDescriptionLength = 4096;
 
         /// <summary>
-        /// Returns the maximum total embed length allowed by Discord.
+        /// Gets the maximum number of fields.
         /// </summary>
-        public const int MaxEmbedLength = 6000;
+        public const int MaxFieldCount = 25;
 
         /// <summary>
-        /// Returns the maximum author name length allowed by Discord.
+        /// Gets the maximum length of a field name.
+        /// </summary>
+        public const int MaxFieldNameLength = 256;
+
+        /// <summary>
+        /// Gets the maximum length of a field value.
+        /// </summary>
+        public const int MaxFieldValueLength = 1024;
+
+        /// <summary>
+        /// Gets the maximum length of an author's name.
         /// </summary>
         public const int MaxAuthorNameLength = 256;
 
         /// <summary>
-        /// Returns the maximum footer length allowed by Discord.
+        /// Gets the maximum length of a footer.
         /// </summary>
         public const int MaxFooterTextLength = 2048;
+
+        /// <summary>
+        /// Gets the maximum overall size of an embed.
+        /// </summary>
+        public const int MaxEmbedLength = 6000;
 
         /// <summary>
         /// The default embed color.
--- a/Remora.Discord.Extensions/Embeds/Constants.cs
+++ b/Remora.Discord.Extensions/Embeds/Constants.cs
@@ -1,0 +1,67 @@
+﻿//
+//  Constants.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) 2017 Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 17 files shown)


Remora.Discord.Extensions/Embeds/Constants.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Returns the maximum title length allowed by Discord.
+        /// </summary>
+        public const int MaxTitleLength = 256;
+
+        /// <summary>
+        /// Returns the maximum description length allowed by Discord.
+        /// </summary>
+        public const int MaxDescriptionLength = 2048;
```
VelvetThePanda, line 45: This was bumped to 4096 last I checked

VelvetThePanda: https://discord.dev/resources/channel#embed-limits-limits states 4096

Remora.Discord.Extensions/Embeds/EmbedBuilder.cs
```diff
@@ -0,0 +1,384 @@
...
+        /// Gets or sets the description of the embed.
+        /// </summary>
+        [MaxLength(Constants.MaxDescriptionLength)]
+        public string Description { get; set; } = string.Empty;
+
+        /// <summary>
+        /// Gets or sets the url of the embed.
+        /// </summary>
+        [Url]
+        public string Url { get; set; } = string.Empty;
```
Nihlus, line 63: If this an Url, should it not be defined as a `Uri`?

Remora.Discord.Extensions/Embeds/EmbedBuilder.cs
```diff
@@ -0,0 +1,384 @@
...
+        /// Adds an <see cref="EmbedAuthor"/> the <see cref="EmbedBuilder"/>.
+        /// </summary>
+        /// <param name="name">The author's name.</param>
+        /// <param name="url">The author's website.</param>
+        /// <param name="iconUrl">The url of the author's icon.</param>
+        /// <returns>The current <see cref="EmbedBuilder"/> for chaining.</returns>
+        public EmbedBuilder WithAuthor
+        (
+            [MaxLength(Constants.MaxAuthorNameLength)] string name,
+            [Url] string url = "",
```
Nihlus, line 254: See previous comment.

Remora.Discord.Extensions/Embeds/EmbedBuilder.cs
```diff
@@ -0,0 +1,384 @@
...
+            [Url] string iconUrl = ""
```
Nihlus, line 255: See previous comment.

Remora.Discord.Extensions/Embeds/EmbedBuilder.cs
```diff
@@ -0,0 +1,384 @@
...
+{
+    /// <summary>
+    /// Provides utilities for building an embed.
+    /// </summary>
+    public class EmbedBuilder
+    {
+        /// <summary>
+        /// Gets or sets the title of the embed.
+        /// </summary>
+        [MaxLength(Constants.MaxTitleLength)]
```
Nihlus, line 45: I'm unfamiliar with these annotations, so forgive me if I ask some obvious questions x)

Is there some analyzer or other mechanism that checks the validity of input or output data from members annotated with these attributes? If not, what's the use case for these over <remarks> in the doc header?

Nihlus (changes requested):

Remora.Discord.Extensions/Embeds/EmbedBuilder.cs
Foxtrek64: These annotations live in the `System.ComponentModel.DataAnnotations` library. It looks like they were originally written for ASP.NET but were ported to be more generic with the advent of .NET Core. Either way, the analyzer in this case appears to be the .NET Runtime itself (through `System.Diagnostics.CodeAnalysis`). I'm not entirely familiar with the back-end system, but `MaxLengthAttribute` has an [`IsValid()`](https://source.dot.net/#System.ComponentModel.Annotations/System/ComponentModel/DataAnnotations/MaxLengthAttribute.cs,0e085bd95ad8e4c6,references) method which returns a boolean. I believe by default the system then throws an IOE or ArgumentOutOfRangeException.

Remora.Discord.Extensions/Embeds/EmbedBuilder.cs
Foxtrek64: This field is a string because [`IEmbed.Url`](https://github.com/Nihlus/Remora.Discord/blob/master/Backend/Remora.Discord.API.Abstractions/API/Objects/Messages/IEmbed.cs#L55) is a string. The `UrlAttribute` ensures that it is a properly formatted url.

Nihlus: Gotcha.

VelvetThePanda (approved):
LGTM

:P_:

Remora.Discord.Extensions/Embeds/EmbedBuilder.cs
Foxtrek64: Added code to the `Ensure()` method to ensure these function.

Nihlus:
I think I'd prefer that the annotations be removed, since they don't appear to provide any tangible differences before runtime (and even then, only as opt-in behaviour). Just documenting the limits in the docheaders will do fine.

Remora.Discord.Extensions/Remora.Discord.Extensions.csproj
```diff
@@ -0,0 +1,29 @@
...
+		<Version>1.0.0</Version>
+		<Description>
+			$(BaseDescription)
+
+			This package provides utilities and components which extened upon base resources.
+		</Description>
+		<PackageReleaseNotes>
+			Initial release.
+		</PackageReleaseNotes>
+		<PlatformTarget>AnyCPU</PlatformTarget>
```
Nihlus, line 17: No need for this property to be set.

Remora.Discord.Extensions/Remora.Discord.Extensions.csproj
```diff
@@ -0,0 +1,29 @@
...
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Remora.Results" Version="7.0.3" />
+		<PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
```
Nihlus, line 22: Remove this reference.

Nihlus (changes requested):

PR #131, opened 2021-12-04
[WIP] Add Voice Implementation (Send only)

carlst99:
This PR hopes to introduce a voice client implementation. Only the sending of voice data is supported; all going well I plan on implementing receiving at a later stage within `Remora.Discord.Unstable`.

I'd love any early feedback y'all can provide, particularly on the exposed API. Otherwise, it is heavily work-in-progress:

- The voice gateway layer is temporary, with no event dispatching and used for testing only. A common gateway layer is in the works.
- There are no unit tests for any voice logic.
- I still need to implement gateway reconnection and UDP heartbeating.
- Switching channels without reconnecting is not yet supported.
- I'm considering attempting a managed encryption implementation.

The `Voice` sample should get you up to speed on how it all comes together. If you're running it on linux, you'll first need to install `libopus-dev` from your package manager.

A broad overview of sending voice data:

- Retrieve a `DiscordVoiceClient` instance for a particular guild from the `DiscordVoiceClientFactory`.
- Call `RunAsync` if the client is not already connected. The client manages its own lifecycle from here, until you call `StopAsync`.
- Submit a stream of audio data. The default transcoder expects PCM-16 (signed, little endian) audio.

Note that I've made minor changes to the voice objects:

- Targeted the concrete object in the `VoiceHeartbeatConverter`
- Removed the `OperationCode` property from `IVoicePayload`. This makes it consistent with the gateway `IPayload` type.


Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
```diff
@@ -0,0 +1,1258 @@
...
+        /// This task will not complete until cancelled (or faulted), maintaining the connection for the duration of it.
+        ///
+        /// If the gateway client encounters a fatal problem during the execution of this task, it will return with a
+        /// failed result. If a shutdown is requested, it will gracefully terminate the connection and return a
+        /// successful result.
+        /// </para>
+        /// </summary>
+        /// <param name="guildID">The ID of the guild to connect to.</param>
+        /// <param name="channelID">The ID of the channel to connect to.</param>
+        /// <param name="isSelfMuted">A value indicating whether the bot should mute itself.</param>
```
VelvetThePanda, line 175: Perhaps I'm nitpicking here, but shouldn't this be along the lines of `Whether the user joins in a muted state`? 

`[...] in a self-muted state` would also work, but perhaps would be a bit long for the documentation.

Also, given this context, I think `joinSelfMuted` would be a more apt parameter name, given what the documentation says it does.

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
```diff
@@ -0,0 +1,1258 @@
...
+        /// Transmits audio.
+        /// </summary>
+        /// <param name="audioStream">The audio data to send.</param>
+        /// <param name="speakingFlags">The speaking flags to use.</param>
+        /// <param name="transcoder">The audio transcoder to use.</param>
+        /// <param name="ct">A <see cref="CancellationToken"/> that can be used to stop the operation.</param>
+        /// <returns>A result representing the outcome of the operation.</returns>
+        public async Task<Result> TransmitAudioAsync
+        (
+            Stream audioStream,
```
VelvetThePanda, line 284: So, if I'm understanding correctly, essentially how this is to be used is you ➜ stream, and so long as you copy data to the stream that's passed, it will continue to play?

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
```diff
@@ -0,0 +1,1258 @@
...
+
+                while (!ct.IsCancellationRequested)
+                {
+                    var read = await audioStream.ReadAsync
+                    (
+                        inputBuffer.Memory[0..sampleSize],
+                        ct
+                    ).ConfigureAwait(false);
+
+                    if (read != sampleSize)
```
VelvetThePanda, line 351: Guess this answers my above question.

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
```diff
@@ -0,0 +1,1258 @@
...
+                    {
+                        break; // TODO: Does this cut off audio in some cases?
```
VelvetThePanda, line 353: To answer the question in this TODO, I'm no expert in audio, but my answer is it depends?

If you're getting the audio from something like a remote server, if we're read to the end of the stream, but the chunk doesn't align with a sample size (which, not sure how likely that is to happen), then there's suddenly no more audio because the server didn't split audio chunks in an apt way.

I *think* the better solution (mind you, I'm an idiot to take this with a bottle of salt) would to be check if the read is 0, which indicates that the stream is empty, and probably finished, though I suppose the cancelation token could also serve that purpose, if an end user is monitoring how long they're streaming audio.

This also brings up the issue of pausing, because a stream might get cut short, but the user might want to continue at a later point in time, which is generally fine, but will also require re-calling this method, so ehhh maybe I'm just being pessimistic.

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
```diff
@@ -0,0 +1,1258 @@
...
+                    }
+
+                    var encodeResult = await transcoder.EncodeAsync
+                    (
+                        inputBuffer.Memory[..read],
+                        opusBuffer.Memory[..read],
```
VelvetThePanda, line 359: In addendum to the above, won't this throw an `IndexOutOfBoundsException`? You'd be off by 1, I think.

On top of this, wouldn't this be the same as the sample size anyway, since you `break` from the loop if it's not the same?

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
```diff
@@ -0,0 +1,1258 @@
...
+            _disconnectRequestedSource.Dispose();
+            _disconnectRequestedSource = new CancellationTokenSource();
+
+            _payloadsToSend.Clear();
+            _receivedPayloads.Clear();
+            _connectionRequestParameters = null;
+            _gatewayConnectionDetails = null;
+            _voiceServerConnectionDetails = null;
+
+            this.ConnectionStatus = GatewayConnectionStatus.Offline;
```
VelvetThePanda, line 486: So, I was gonna comment on previous lines about setting the connection status, but I see it here too.

Is there a reason for setting to offline (In ConnectAsync I believe)?

VelvetThePanda: Actually I may have seen `Disconnected`, correct me if the above comment is wrong.

VelvetThePanda (changes requested):
This is objectively an enormously chonky PR, and given my inexperience with low-level audio stuff in general, my feedback is limited already, but I left a few questions and comments.

Great stuff, otherwise.
heart

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
carlst99: I prefer that description, thanks. I've gone with the below for now.
> A value indicating whether the bot should connect in a muted state.

Given that it's a persistent state till changed, and not just for the period of joining the call, the current name makes more sense to me though.

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
carlst99: Yeah, as you've discovered. Data is just read from the stream in chunks of the size that the transcoder requests. And you can call Transmit multiple times in a row with different streams.

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
carlst99: It's a good point. You have to rely on the stream being 'whole' before dispatching it. I'm considering re-writing this using `System.IO.Pipelines`; it would make situations like this easier to manage as the caller can provide data throughout the process, and the callee can read its desired chunks, waiting until the pipe is closed.

Another thing I'll play around with is filling out the remainder of a sample with zeroes, if the amount of bytes read from the stream is less than expected. This would let me utilise your idea of finishing when zero bytes are read; I naively terminate as I do currently because Opus expects a full sample.

I think pausing should be managed by the caller. There's no issue in re-calling `Transmit..` that I can see. Definitely open to discussion though.

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
carlst99: Well, it hasn't yet, which I take as a good sign 😛. The end value is exclusive.

Yes, you're absolutely correct that it's the sample size. I could use either or, but if for whatever reason down the line we decide to transcode just the data that's been read regardless of sample size, it saves a whopping two seconds of code editing 😂.

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
carlst99: I'm not sure I entirely follow sorry. The `ConnectionStatus` property is used for internal connection control; if we're offline we need to open a new connection and if we're disconnected we need to resume.
The property isn't updated at all in `ConnectAsync` and the instance you've highlighted is in the `CleanupAsync` method, which we perform on shutdown.

AraHaan:
Can we get this into an optional package named ``Remora.Discord.Voice`` and then have it merged and shipped?

I would at least like having send first, then worry about receive later for implementing bot faxing other bot later.

AraHaan:
After looking at the files, could perhaps libopus be downloaded on demand at runtime based upon these factors:
- if the process is running under x86 download x86 opus
- if the process is x64 download x64 opus
- if arm or arm64 use those by downloading them (idk how to check for those 2 yet)
- if windows, use dll version (can easily be checked by the ``OperatingSystem`` class), if Linux use ``libopus.so`` version, if MacOS use ``libopus.dylib`` version.

carlst99:
Linking the `libopus` library certainly needs spiffing up but I don't plan of having it downloaded at runtime. I believe there's more factors in the mix that could go wrong; issues like consumers with extremely tight firewalls come to mind. More importantly, it would require a reliable place to host the binaries.

My current approach is to package all relevant runtime variants into a NuGet package, perhaps named `Remora.Natives`. It will ensure that the correct version of the library is always there as soon as you restore or publish.

AraHaan:
> Linking the `libopus` library certainly needs spiffing up but I don't plan of having it downloaded at runtime. I believe there's more factors in the mix that could go wrong; issues like consumers with extremely tight firewalls come to mind. More importantly, it would require a reliable place to host the binaries.
> 
> My current approach is to package all relevant runtime variants into a NuGet package, perhaps named `Remora.Natives`. It will ensure that the correct version of the library is always there as soon as you restore or publish.

Or ``Remora.Discord.Opus.*RID*`` where rid could be:
- win-x86
- win-x64
- win-arm64 (win-arm no longer exists or is supported by .NET on Windows)
- linux-x64 (x86 linux is no longer supported by .NET since 5?)
- linux-arm
- linux-arm64
- osx-x64
- osx-arm64

Backend/Remora.Discord.Voice/DiscordVoiceClient.cs
```diff
@@ -673,7 +669,7 @@ CancellationToken ct
 
                 _receivedPayloads.Enqueue(sessionDescription);
                 _dataService.Initialize(sessionDescription.Data.SecretKey);
-                SendSilenceFrames(_defaultTranscoder.SampleSize, ct); // Can be the key to enabling voice receiving in some cases
+                SendSilenceFramesAsync(_defaultTranscoder.SampleSize, ct); // Can be the key to enabling voice receiving in some cases
```
Nihlus, line 672: Unobserved task?

Nihlus:
I would prefer it if we did not need to bundle a native libopus on any platform. Is there a managed implementation we can use instead?

AraHaan:
wait why was this closed?

VelvetThePanda:
@AraHaan see [this message](https://canary.discord.com/channels/789912892426027039/789913059116187658/953384661139214387) on the Discord server

PR #132, opened 2021-12-04
Add `IMessageUpdate` to EarlyCacheResponder

VelvetThePanda:
Adds `IResponder<IMessageUpdate>` to the early cache responders; correct me if I did something wrong.

### DIFF
--- a/Backend/Remora.Discord.Caching/Responders/EarlyCacheResponder.cs
+++ b/Backend/Remora.Discord.Caching/Responders/EarlyCacheResponder.cs
@@ -49,6 +49,7 @@ namespace Remora.Discord.Caching.Responders
         IResponder<IGuildRoleCreate>,
         IResponder<IGuildRoleUpdate>,
         IResponder<IMessageCreate>,
+        IResponder<IMessageUpdate>,
         IResponder<IMessageReactionAdd>,
         IResponder<IUserUpdate>,
         IResponder<IInteractionCreate>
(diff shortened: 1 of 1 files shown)


Backend/Remora.Discord.Caching/Responders/EarlyCacheResponder.cs
```diff
@@ -238,6 +239,25 @@ public Task<Result> RespondAsync(IMessageCreate gatewayEvent, CancellationToken
...
+                return Task.FromResult(Result.FromSuccess());
+            }
+
+            if (!gatewayEvent.ID.IsDefined(out var messageID))
+            {
+                return Task.FromResult(Result.FromSuccess());
+            }
+
+            var key = KeyHelpers.CreateMessageCacheKey(channelID, messageID);
+            _cacheService.Cache(key, gatewayEvent);
```
Rutherther, line 256: It's probably better to specify the generic argument `Cache<IMessage>`, because the logic of the Cache method will also try to store the referenced message that way

VelvetThePanda: Well, the event only exposes the data about the message, and not the message itself; I could however add logic to attempt to cache the referenced message, though? If that works better I mean

Rutherther: Wait I didn't notice that the event only has `IPartialMessage`, but then your code probably won't work. When you call `IDiscordRestChannelAPI.GetChannelMessageAsync`, it will try to get `IMessage` from the cache only. So you will have to try to get the value from cache and update it with values you have I think. And if it isn't there you will probably have to create whole `IMessage` anyway and fill it with data you have.

Rutherther: When I am looking at guild member update, it looks like if the previous value was not stored in the cache, then the new one is just thrown away, not stored anywhere. That makes sense, but is also kinda bad at you won't have the data stored. So I am not really sure how to resolve the case when you don't have the message in cache already and you won't be able to fill the data that are required, if you won't have them.

VelvetThePanda: Hmm. That is quite the predicament.

Nihlus:
As discussed on the server, this PR won't function correctly in its current state, and will actually break the caching implementation (since it expected an `IMessage`, not an `IPartialMessage` to be in the cache). 

I'm closing this for now, and we'll have to get back to it once we figure out how to do this properly.

PR #133, opened 2021-12-05
Fix channel caching on `GUILD_CREATE`

VelvetThePanda:
Fixed an issue where channels were being cached under the guild ID, causing cache to be constantly overwritten, and channels to not be cached correctly.

I also added parentheses to the type check, as it would cause an edge case where group DMs would be cached "incorrectly"; at least, I can't imagine that was inentional either.


Nihlus:
Nice catch, thanks.

Issue #119, opened 2021-10-13
Occasionally receiving "The payload after identification was not a Ready payload." on reconnect


Cyberboss:
Still working on a reproduction for this, but rapidly disconnecting then reconnecting a bot (full service provider teardown and rebuild), seems to be able to trigger this error case here:

https://github.com/Nihlus/Remora.Discord/blob/85385a82e2e3a73aa4a86753bd0396505733b009/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs#L829-L836

My case is 2-4 bots using the same token running an integration test simultaneously in CI.

Foxtrek64:
I'll add a change to the logging here to have it post what kind of event it actually receives and have it dump the raw output of that to the debug log.

Hopefully that tells us exactly what's happening.

Foxtrek64:
So we have some proper test data:

Iteration 1 is not problematic:

```
info: Remora.Discord.ReadyPayload.Program[0]
      Iteration: 1
info: Remora.Discord.ReadyPayload.DiscordProvider[0]
      Gateway ready. Version 9
info: Remora.Discord.ReadyPayload.DiscordProvider[0]
      Stoptting gateway client.
warn: Remora.Discord.ReadyPayload.DiscordProvider[0]
      Gateway issue: The transport service is not connected.
```

The warning is there because I think even though `gatewayClient.RunAsync()` is called, it's not given enough time to run before it is aborted, so it never ends up actually connecting. Iteration 2 is where it gets interesting:

```
info: Remora.Discord.ReadyPayload.Program[0]
      Iteration: 2
info: Remora.Discord.ReadyPayload.DiscordProvider[0]
      Disconnect detected
info: Remora.Discord.ReadyPayload.DiscordProvider[0]
      Stoptting gateway client.
warn: Remora.Discord.ReadyPayload.DiscordProvider[0]
      Gateway issue: The payload after identification was not a Ready payload.
        Expected: Remora.Discord.API.Abstractions.Gateway.IPayload`1[[Remora.Discord.API.Abstractions.Gateway.Events.IReady, Remora.Discord.API.Abstractions, Version=53.0.0.0, Culture=neutral, PublicKeyToken=null]]
        Actual: Remora.Discord.API.Payload`1[[Remora.Discord.API.Abstractions.Gateway.Events.IInvalidSession, Remora.Discord.API.Abstractions, Version=53.0.0.0, Culture=neutral, PublicKeyToken=null]]
```

[...]

PR #151, opened 2022-01-21
Implement better logging for payload

Foxtrek64:
This PR updates Remora.Discord.Gateway to a beta version as it includes code useful for diagnosing #119 but that is not suitable for a full release.


Foxtrek64:
Some unrelated changes on a different branch got pulled in because apparently changing branches no longer automatically caches changes. I've cleaned those up.

My plan with this is to have us release the beta version here, then run the test as described in #119 and reproduce the error message. Then we can identify what's actually being sent and add a case for it.

Nihlus:
I'm fine having this as a permanent change - it's useful info either way.

PR #159, opened 2022-02-26
[WIP] Initial implementation of localized application commands.

Nihlus:
This PR implements support for localized slash commands (and friends) using GNU gettext.

To enable localized commands, call `.AddLocalizations` on your service collection. To provide the actual localizations themselves, bundle them as embedded resources in your entry assembly under a folder named `remora.locales` - `AddLocalizations` should then pick them up.

The assembly the localizations are loaded from can be customized by passing it to `AddLocalizations`, as can the name of the containing folder. Localization files are expected to be named the same as their matching locale, and to have the `.mo` file ending (`sv-SE.mo`, for example).

## TODO
  - [ ] Additive localizations from multiple assemblies
  - [ ] Actual testing against Discord API; I'm not in the beta
  - [ ] Localizations in places like message components or modals
  - [ ] Tests


VelvetThePanda (approved):

AraHaan:
Why not use the built in localization functionality in .NET from ``System.Resources.ResourceManager``?

Nihlus:
@AraHaan That system generally uses satellite assemblies, and is not as universally used as `.po` and `.mo` files. Using gettext enables the use of far more tooling across the translation ecosystem.

AraHaan:
My old translation system back when I done my 1st bot with Discord.py was to load them from json. I guess that could also be an option since using System.Text.Json is cheap as well.

Nihlus:
I don't intend on supporting more than one system. It adds needless complexity. .po and .mo files are industry standard, not just in C#'s ecosystem.

VelvetThePanda (approved):
Took a onceover since this is somewhat out of my field; looks *fine*; I'm not very competent in the ways of translation, so I can't say much about how parsing is handled, but the API looks OK to me I think?

PR #160, opened 2022-03-01
System.Text.Json on .NET 6 should come from runtime pack.

AraHaan:
S.T.J (since stable 6.0.2 is referenced here) should be ignored when targeting .NET 6. This is due to it being included in the Microsoft.NETCore.App runtime bundle which often times includes crucial bug fixes / improvements. As such if newer versions is needed, simply ensure the user's .NET SDK is updated on their end or have them do a property to override the default runtime pack version provided by the SDK).

Depends on:
- https://github.com/azambrano/JsonDocumentPath/pull/2
- https://github.com/Nihlus/Remora.Rest/pull/4

### DIFF
--- a/Backend/Remora.Discord.API/Remora.Discord.API.csproj
+++ b/Backend/Remora.Discord.API/Remora.Discord.API.csproj
@@ -21,7 +21,7 @@
     <ItemGroup>
         <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
         <PackageReference Include="Microsoft.Extensions.Options" Version="6.0.0" />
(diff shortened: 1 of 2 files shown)


Nihlus:
I'm honestly not sure about this one; the added condition increases maintenance load, complicates things in the future, and doesn't really add much beyond avoiding a few kilobytes of downloaded assemblies on restore.

The security argument doesn't really hold any water either, since the same patches would be released onto nuget at the same or earlier time as the runtime would have them.

I'm closing this PR for now, but feel free to add additional reasons; if I change my mind later, we can reopen.

PR #162, opened 2022-03-02
Fix UB when caching derivative entity types

VelvetThePanda:
Fixes an issue where caching a derivative entity type would cache under the subtype instead of the expected super type.

An example of this is caching `IGuild` from the `IGuildCreate` event because despite a switch expresison matching `IGuild`, the compiler (and runtime for that matter) see `IGuildCreate` (or it's implementation). When the generic is inferred for `CacheInstance` the compiler/runtime will choose the derivative (IGuildCreate) instead of the base (IGuild).

I believe tests should be written for the cache service to help prevent transient bugs like this

### DIFF
--- a/Backend/Remora.Discord.Caching/Services/CacheService.cs
+++ b/Backend/Remora.Discord.Caching/Services/CacheService.cs
@@ -131,7 +131,7 @@ public class CacheService
 
     private void CacheWebhook(object key, IWebhook webhook)
     {
-        CacheInstance(key, webhook);
+        CacheInstance<IWebhook>(key, webhook);
 
         if (!webhook.User.IsDefined(out var user))
         {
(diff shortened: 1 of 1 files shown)


Nihlus:
These changes do nothing; the compiler infers the type from the compile-time parameter type, not the runtime type. If you're seeing an issue where `IGuildCreate` is not creating a cache entry, it's almost assuredly due to something else.

VelvetThePanda:
Appears that this seems to be an issue with inconsistencies that I can't seem to reproduce when using an IMemoryCache; it also appears that I may have been negligent to the nuance in type when I was attempting to cache presences. Either way I'm not sure where this issue resides (if it even exists, I may just be stupid). I'll try to do more testing and validation before hastily opening a PR next time.

PR #165, opened 2022-03-10
Fix unexpected/undocumented behavior when setting absolute expirations

VelvetThePanda:
This is a *proper* fix, unlike what #162 was, since I've taken some time to properly investigate the root issue with this. While this is technically probably documented behavior in the BCL, both Remora's documentation and smaples are technically misleading (which is more than likely unintentionally so).

This issue is not a fault of Remora, but of deeply nested behavior in the BCL, and specifically in CacheEntry.

the tl;dr is essenitally if a sliding expiration is set, and absolute is *not*, the sliding expiration will take precedence over the absolute expiration. Joyous. 

As I said, I've done proper research into this. A link to culprit BCL code is [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Caching.Memory/src/CacheEntry.cs#L181-L207)

Example screenshot of before this fix:

image_1

And after:

image_2

A bandaid fix would be to simply use `.SetDefaultSlidingExpiration` and set it to null, but I think this is a better fix as it aligns more to what people probably expect `.SetAbsoluteExpiration` does. Sliding expirations are strange and exhibit strange behavior. I'm quite against them.

### DIFF
--- a/Backend/Remora.Discord.Caching/Services/CacheSettings.cs
+++ b/Backend/Remora.Discord.Caching/Services/CacheSettings.cs
@@ -183,7 +183,9 @@ public class CacheSettings
         }
 
         var slidingExpiration = GetSlidingExpirationOrDefault<T>();
-        if (slidingExpiration is not null)
+
+        // See https://github.com/Nihlus/Remora.Discord/pull/165
+        if (slidingExpiration is not null && absoluteExpiration is not null)
         {
             cacheOptions.SetSlidingExpiration(slidingExpiration.Value);
         }


VelvetThePanda:
Why must Git suck

VelvetThePanda:
I opened an issue on the runtime repo regarding this, and hope the MSDN documentation can be updated to reflect this nuance. 

https://github.com/dotnet/runtime/issues/66445

Nihlus:
Should users not be able to set a sliding expiration only? The two are independently settable because of this behaviour, and setting one or the other should be the same as manipulating the cache options directly.

VelvetThePanda:
I suppose so; as I stated this is moreso to align with what could be considered "Absolute" and "Sliding" expirations, as it's not very clearly documented that a sliding expiration will take precedence over the absolute expiration if the message isn't touched. 

I had more written out here, but I went through the caching example, and I may have proven myself to be incompetent at doing so much as reading twice now.

PR #166, opened 2022-03-13
Redis-backed caching

VelvetThePanda:
This PR aims to implement the same basic concept that #58 did, and as such that PR provides. As a result, this PR is at least inspired by that PR, though everything was rewritten from scratch.

This PR should be simplier than the former, as well, ditching a complex cache key struct in favor of rewriting `KeyHelpers` to return colon-delimited strings, which should be organized in a topological manner akin to how they relate to parent entities (e.g. `Channel:{channelID}:Message:{messageID}`).

This is a large PR with lots of moving parts, so any feedback on potential bugs, weird wording, etc is appreciated!


AraHaan:
I wonder if there should be a feature to have cache be done to an sqlite db if we wanted.

Foxtrek64:
> I wonder if there should be a feature to have cache be done to an local sqlite db if we wanted.

Agreed. Whatever implementation we apply here should be done in a way that's store-agnostic. That is, we should be able to use MemoryCache, redis, sqlite, or any other solution simply by providing a service that can write to that store.

VelvetThePanda:
I personally find saving the idea of saving potentially tens of thousands of entities in a database like that to be a bit strange, as well as a violation of GDPR. However, since the cache was abstracted to an interface, you're welcome to replace the implementation either way (`IServiceCollection.Replace`), or even add your own impl before calling this method.

VelvetThePanda:
I also forgot to mention that the memory cache IS the default option (this PR does not change default behavior of the caching package). There is also already a default implementation for caching in Redis. As this is already somewhat niche, and SQLite would probably be even more niche, I'm not personally sure adding another dependency/extension method/implementation very very few people would use is worth it, and it's probably better to implement it yourself in case you decide to change the backing store in some way

Nihlus:
This looks fine, but Redis caching should live in its own assembly. There's no need to pull down that as a dependency for applications that don't use it.

VelvetThePanda:
If the redis package were to be placed in it's own assembly, I take it would have a dependency on the original caching package then? Also would it be preferred to make the method that adds the caching APIs public?

VelvetThePanda:
Both methods to enable caching will implicitly enable caching behaviours (caching APIs, responders);

`IServiceCollection.AddDiscordCaching()` and `AddDiscordRedisCaching()` respectively. 

In terms of configuring the connection options for Redis (say, running in Docker, or perhaps on Azure), a configuration action can be passed to the latter method, expanding it out to something along the lines of this;

```cs
// configuration could come from anywhere, but likely provided by the generic host.

var redisConfiguration = configuration.GetConnectionString("redis");

services.AddDiscordRedisCaching(redis => redis.Configuration = redisConfiguration);
```

VelvetThePanda:
I forgot to note that a user can also choose to only add the caching APIs and provide their own implementation of `ICacheService` if they please (I believe someone here mentioned caching to SQLite) by calling `.AddCachingAPIAndRespodners()` on the service collection. There's a note in the documentation that a caching implementation is required if that method is called directly.

Nihlus:
Sounds good!

Backend/Remora.Discord.Caching.Redis/Services/RedisCacheService.cs
```diff
@@ -0,0 +1,353 @@
...
+            return;
+        }
+
+        var evictionKey = $"Evicted:{key}";
+        var settings = _cacheSettings.GetRedisEntryOptions<TInstance>();
+
+        await _cache.SetAsync(evictionKey, value, settings);
+    }
+
+    private async Task CacheWebhook(string key, IWebhook webhook)
```
Nihlus, line 139: Task-return methods should be suffixed with `Async`.

Nihlus, line 139: Could we break these out into a shared implementation that uses the cache service under the hood? It's a little silly to have exactly the same code in two places.

Backend/Remora.Discord.Caching/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -48,16 +50,33 @@ public static class ServiceCollectionExtensions
...
+    }
+
+    /// <summary>
+    /// Adds a caching implementations of various Discord API types and caching responders to the service collection.
+    /// </summary>
+    /// <param name="services">The service collection to add the caching APIs to.</param>
+    /// <remarks>Invoking this method alone is not recommended as an implementation of
+    /// <inheritdoc cref="ICacheService"/> must also be provided. Consider using <see cref="AddDiscordCaching"/> instead.</remarks>
+    /// <returns>The service collection to chain calls with.</returns>
+    public static IServiceCollection AddCachingAPIAndResponders(this IServiceCollection services)
```
Nihlus, line 77: Move the responder adding into `AddDiscordCaching` (since those always need to be included when caching is enabled) and rename this to just `AddCachingAPI`. There could probably be a case made for not having this method at all, and just having it all in `AddDiscordCaching`, too.

Nihlus (changes requested):

Backend/Remora.Discord.Caching/Extensions/ServiceCollectionExtensions.cs
VelvetThePanda: Well I considered this, but `AddDiscordRedisCaching` requires them too, but then you have a potentially needless `IMemoryCache` added to your container. I'd be fine with reverting this change though.

Backend/Remora.Discord.Caching/Services/ICacheProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// <param name="absoluteExpiration">The absolute expiration of the value to cache.</param>
+    /// <param name="slidingExpiration">The sliding expiration of the value to cache.</param>
+    /// <param name="ct">A cancellation token to cancel the operation.</param>
+    /// <typeparam name="TInstance">The type to cache.</typeparam>
+    /// <remarks>Absolute and sliding expirations may be handled differently (or not at all)
+    /// depending on the implementation, and are simply a hint to the implementation that the
+    /// value should have a pre-determined lifetime in it's backing-store.</remarks>
+    /// <returns>A <see cref="ValueTask"/> representing the result of the potentially asynchronous operation.</returns>
+    ValueTask CacheAsync<TInstance>
+        (string key, TInstance instance, TimeSpan? absoluteExpiration = null, TimeSpan? slidingExpiration = null, CancellationToken ct = default)
```
Nihlus, line 49: Fix formatting here.

Backend/Remora.Discord.Caching/Services/MemoryCacheProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// </summary>
+    /// <param name="memoryCache">The memory cache.</param>
+    public MemoryCacheProvider(IMemoryCache memoryCache)
+    {
+        _memoryCache = memoryCache;
+    }
+
+    /// <inheritdoc cref="ICacheProvider.CacheAsync{TInstance}" />
+    public ValueTask CacheAsync<TInstance>
+        (string key, TInstance instance, TimeSpan? absoluteExpiration = null, TimeSpan? slidingExpiration = null, CancellationToken ct = default)
```
Nihlus, line 50: Fix formatting here.

Samples/Caching/Caching.csproj
```diff
@@ -11,6 +11,7 @@
     </ItemGroup>
 
     <ItemGroup>
+        <ProjectReference Include="..\..\Backend\Remora.Discord.Caching.Redis\Remora.Discord.Caching.Redis.csproj" />
```
Nihlus, line 14: Is there a reason the caching sample now uses Redis instead of normal memory caching? A separate redis project might be more appropriate.

Backend/Remora.Discord.Caching.Redis/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <returns>The services, with caching enabled.</returns>
+    public static IServiceCollection AddDiscordRedisCaching(this IServiceCollection services, Action<RedisCacheOptions>? configureRedisAction = null)
+    {
+        configureRedisAction ??= (s) => s.ConfigurationOptions = new ConfigurationOptions
+        {
+            EndPoints = { { "localhost", 6379 } }
+        };
+
+        services.AddDiscordCaching();
+        services.AddStackExchangeRedisCache(configureRedisAction);
```
Nihlus, line 61: Does Redis support other types of caches? If so, it might be inappropriate to assume that the user wants a stack-exchange cache here. This question mainly stems from me having zero experience with Redis, though.

Backend/Remora.Discord.Caching.Redis/Services/RedisCacheProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+    public async ValueTask CacheAsync<TInstance>
+        (string key, TInstance instance, TimeSpan? absoluteExpiration = null, TimeSpan? slidingExpiration = null, CancellationToken ct = default)
+        where TInstance : class
+    {
+        if (absoluteExpiration == TimeSpan.Zero)
+        {
+            return;
+        }
+
+        var serialized = JsonSerializer.Serialize(instance, _jsonOptions);
```
Nihlus, line 67: What happens if the type is not trivially serializable or is otherwise configured for JSON serialization here? This works fine for Discord entities, but in the case of more complex objects where the memory cache handles everything users might end up with unexpected behaviour.

Backend/Remora.Discord.Caching/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -48,17 +50,20 @@ public static class ServiceCollectionExtensions
     /// <remarks>
     /// The cache uses a run-of-the-mill <see cref="IMemoryCache"/>. Cache entry options for any cached type can be
     /// configured using <see cref="IOptions{CacheSettings}"/>.
+    ///
+    /// When choosing a cache implementation, it should be noted that <see cref="AddDiscordCaching"/> and
```
Nihlus, line 54: Is it not more appropriate to always require `AddDiscordCaching`, and have `AddDiscordRedisCaching` as an "addon" to the base caching services?

Backend/Remora.Discord.Caching/Responders/EarlyCacheResponder.cs
```diff
@@ -158,19 +158,23 @@ public Task<Result> RespondAsync(IGuildMembersChunk gatewayEvent, CancellationTo
...
-    public Task<Result> RespondAsync(IGuildMemberUpdate gatewayEvent, CancellationToken ct = default)
+    public async Task<Result> RespondAsync(IGuildMemberUpdate gatewayEvent, CancellationToken ct = default)
     {
         var key = KeyHelpers.CreateGuildMemberKey(gatewayEvent.GuildID, gatewayEvent.User.ID);
 
         // Since this event isn't playing nice, we'll have to update by creating an object of our own.
-        if (_cacheService.TryGetValue<IGuildMember>(key, out var cachedInstance))
+        var cacheResult = await _cacheService.TryGetValueAsync<IGuildMember>(key);
+
+        IGuildMember cachedInstance = cacheResult.Entity;
```
Nihlus, line 175: Use `var`.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
         if (_cacheSettings.GetAbsoluteExpirationOrDefault(typeof(TInstance)) is var absoluteExpiration)
         {
             if (absoluteExpiration == TimeSpan.Zero)
             {
-                return;
+                return default;
             }
         }
 
         Action cacheAction = instance switch
```
Nihlus, line 76: Rework this so it handles asynchronous execution properly.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
     /// <typeparam name="TInstance">The type of the value.</typeparam>
     /// <param name="key">The cache key.</param>
-    public void Evict<TInstance>(object key)
+    /// <param name="ct">A cancellation token to cancel the operation.</param>
+    /// <returns>A <see cref="ValueTask"/> representing the potentially asynchronous operation.</returns>
+    public async ValueTask<Result<TInstance>> EvictAsync<TInstance>(string key, CancellationToken ct = default)
         where TInstance : class
     {
-        if (_memoryCache.TryGetValue(key, out var existing))
+        var cacheResult = await _cacheProvider.EvictAsync<TInstance>(key, ct);
```
Nihlus, line 126: `evictionResult` might be a more appropriate name here.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
+        await _cacheProvider.CacheAsync(key, cacheResult.Entity, options.AbsoluteExpirationRelativeToNow, options.SlidingExpiration, ct);
+
+        return cacheResult.Entity;
     }
 
-    private void CacheWebhook(object key, IWebhook webhook)
+    private async Task CacheWebhookAsync(string key, IWebhook webhook)
     {
-        CacheInstance(key, webhook);
+        CacheInstanceAsync(key, webhook);
```
Nihlus, line 142: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
         var userKey = KeyHelpers.CreateUserCacheKey(user.ID);
-        Cache(userKey, user);
+        await CacheAsync(userKey, user);
     }
 
-    private void CacheTemplate(object key, ITemplate template)
+    private async Task CacheTemplateAsync(string key, ITemplate template)
     {
-        CacheInstance(key, template);
+        CacheInstanceAsync(key, template);
```
Nihlus, line 155: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
         var creatorKey = KeyHelpers.CreateUserCacheKey(template.Creator.ID);
-        Cache(creatorKey, template.Creator);
+        await CacheAsync(creatorKey, template.Creator);
     }
 
-    private void CacheIntegration(object key, IIntegration integration)
+    private async Task CacheIntegrationAsync(string key, IIntegration integration)
     {
-        CacheInstance(key, integration);
+        CacheInstanceAsync(key, integration);
```
Nihlus, line 163: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
         var userKey = KeyHelpers.CreateUserCacheKey(user.ID);
-        Cache(userKey, user);
+        await CacheAsync(userKey, user);
     }
 
-    private void CacheBan(object key, IBan ban)
+    private async Task CacheBanAsync(string key, IBan ban)
     {
-        CacheInstance(key, ban);
+        CacheInstanceAsync(key, ban);
```
Nihlus, line 176: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
         var userKey = KeyHelpers.CreateUserCacheKey(ban.User.ID);
-        Cache(userKey, ban.User);
+        await CacheAsync(userKey, ban.User);
     }
 
-    private void CacheGuildMember(object key, IGuildMember member)
+    private async Task CacheGuildMemberAsync(string key, IGuildMember member)
     {
-        CacheInstance(key, member);
+        CacheInstanceAsync(key, member);
```
Nihlus, line 184: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -64,137 +59,142 @@ IOptions<MemoryCacheOptions> memoryCacheOptions
...
         var userKey = KeyHelpers.CreateUserCacheKey(user.ID);
-        Cache(userKey, user);
+        await CacheAsync(userKey, user);
     }
 
-    private void CacheGuildPreview(object key, IGuildPreview preview)
+    private async Task CacheGuildPreviewAsync(string key, IGuildPreview preview)
     {
-        CacheInstance(key, preview);
+        CacheInstanceAsync(key, preview);
```
Nihlus, line 197: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -204,13 +204,13 @@ private void CacheGuildPreview(object key, IGuildPreview preview)
...
-            Cache(emojiKey, emoji);
+            await CacheAsync(emojiKey, emoji);
         }
     }
 
-    private void CacheGuild(object key, IGuild guild)
+    private async Task CacheGuildAsync(string key, IGuild guild)
     {
-        CacheInstance(key, guild);
+        CacheInstanceAsync(key, guild);
```
Nihlus, line 213: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -258,52 +258,52 @@ private void CacheGuild(object key, IGuild guild)
...
-            Cache(roleKey, role);
+            await CacheAsync(roleKey, role);
         }
     }
 
-    private void CacheEmoji(object key, IEmoji emoji)
+    private async Task CacheEmojiAsync(string key, IEmoji emoji)
     {
-        CacheInstance(key, emoji);
+        CacheInstanceAsync(key, emoji);
```
Nihlus, line 277: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -258,52 +258,52 @@ private void CacheGuild(object key, IGuild guild)
...
         var creatorKey = KeyHelpers.CreateUserCacheKey(creator.ID);
-        Cache(creatorKey, creator);
+        await CacheAsync(creatorKey, creator);
     }
 
-    private void CacheInvite(object key, IInvite invite)
+    private async Task CacheInviteAsync(string key, IInvite invite)
     {
-        CacheInstance(key, invite);
+        CacheInstanceAsync(key, invite);
```
Nihlus, line 290: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -258,52 +258,52 @@ private void CacheGuild(object key, IGuild guild)
...
         var inviterKey = KeyHelpers.CreateUserCacheKey(inviter.ID);
-        Cache(inviterKey, inviter);
+        await CacheAsync(inviterKey, inviter);
     }
 
-    private void CacheMessage(object key, IMessage message)
+    private async Task CacheMessageAsync(string key, IMessage message)
     {
-        CacheInstance(key, message);
+        CacheInstanceAsync(key, message);
```
Nihlus, line 303: Don't leave tasks unobserved.

Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -316,12 +316,12 @@ private void CacheMessage(object key, IMessage message)
...
 
-        Cache(referencedMessageKey, referencedMessage);
+        await CacheAsync(referencedMessageKey, referencedMessage);
     }
 
-    private void CacheChannel(object key, IChannel channel)
+    private async Task CacheChannelAsync(string key, IChannel channel)
     {
-        CacheInstance(key, channel);
+        CacheInstanceAsync(key, channel);
```
Nihlus, line 324: Don't leave tasks unobserved.

Nihlus (changes requested):

Samples/Caching/Caching.csproj
VelvetThePanda: I believe this was a by-product of migrating things to a new project.

Backend/Remora.Discord.Caching.Redis/Extensions/ServiceCollectionExtensions.cs
VelvetThePanda: I take it you're referring to the `StackExchange` part?

VelvetThePanda: Can't read; don't mind me. `StackExchange` is just the name of the company that made the library. These people: https://stackexchange.com/

Backend/Remora.Discord.Caching.Redis/Services/RedisCacheProvider.cs
VelvetThePanda: Did you have any particular example?

Nihlus: In particular, I'm thinking of `RateLimitBucket`, which does not currently have well-defined serialization. Generally, this is a concern because the method signatures on `ICacheProvider` does not imply that the instance that is to be cached must be JSON serializable, and as such users might try to cache any arbitrary object - I certainly would. If we want this to be a requirement, it should be made much clearer, and should have some sort of additional check beyond just doing the serialization and hoping for the best.

Nihlus: Serialization could of course be added for types we control, but that does not neccesarily translate to consumers of the library.

VelvetThePanda: Fair enough. Not sure how you'd even go about checking for such a thing, though.

Backend/Remora.Discord.Caching/Extensions/ServiceCollectionExtensions.cs
VelvetThePanda: The comment is moreso in reference to the behavior exhibited by Remora's caching; if you were to use Redis caching, Remora will always cache to Redis, and vice versa.

Nihlus: Might want to reword this, then.

 Another (somewhat unrelated) thought - having both MemoryCacheProvider and RedisCacheProvider available (regardless of whether you've called AddDiscordRedisCaching) in the service provider might be useful, in the case where you want to cache data strictly in-process.

I'm thinking that instead of calling `Replace`, just adding the redis provider as another instance of `ICacheProvider` (thereby making that one take precedence) might be more appropriate; alternatively, adding both first as their plain types, and then making a separate forwarding call to `AddSingleton<ICacheProvider, ...>`.

Backend/Remora.Discord.Caching/Services/MemoryCacheProvider.cs
VelvetThePanda: Would it be more appropriate to put it on one line? Or break it out even more?..

Backend/Remora.Discord.Caching/Extensions/ServiceCollectionExtensions.cs
VelvetThePanda: Doesn't this cause registration issues when the service provider attempts to resolve it? Or does it simply chose the last added type?

Nihlus: If you register more than one service as a provider of the same interface, the last added type is used if the interface is requested. All services that implement that interface can be requested by using `GetServices`, or the individual services by their concrete types (if they've been registered as their concrete types).

Backend/Remora.Discord.Caching/Services/MemoryCacheProvider.cs
Nihlus: The rest of the library uses Allman-style line breaks for very long lines, regardless of where the construct is; that is, this method should put each parameter on its own line, with the parentheses placed like you normally would place braces.

Nihlus:
Looks good now! We still need to figure out the best way to document / adapt to non-trivially serializable cache entities - as I said in the conversation, requiring cached entities to be JSON-serializable is fine, but it needs to be clearly communicated. 

As soon as we get that done, I think this PR is ready for merge.

VelvetThePanda:
> We still need to figure out the best way to document / adapt to non-trivially serializable cache entities

I've made the methods virtual so people can override this behavior as they please, should JSON not be a suitable method of serialization (e.g. preferring protobuf instead).

> [...] requiring cached entities to be JSON-serializable is fine, but it needs to be clearly communicated.
Documentation has also been added in a footnote on the caching methods, as well as the extension method 👍🏽

Nihlus:
Sounds good! You have some malformed XML warnings to take care of, too.

VelvetThePanda:
@Nihlus thanks for pointing that out; should be good to go now 👍🏽

Nihlus (approved):

PR #170, opened 2022-03-19
Add `MeansImplicitUse` to `Add____` extension methods

VelvetThePanda:
Adds the `[MeansImplicitUse]` attribute to type parameters of extension methods that add user-provided services to the container


Nihlus:
`DynamicallyAccessedMembers` is probably more appropriate, since it improves tree trimming correctness.

PR #176, opened 2022-03-27
Common Gateway Implementation

carlst99:
This PR provides my take on a common gateway infrastructure. As to whether or not it is sensible, I'll leave to you kind reviewers 😄.

Here is a brief overview of the changes.

### The `IGatewayClient`

This defines the interface for a gateway client. This is similar to the public interface of the current `DiscordGatewayClient`, with the addition of the 'pre-shutdown commands' concept. This allows consumers to de/register commands that they wish to be sent immediately before a user-requested shutdown is initiated. The primary use case for this is the voice gateway needing to update its voice status when the program is terminated.

Currently, the `IGatewayClient` is not registered to the service collection.

### The `BaseGatewayClient`

This abstract class handles the sending and receiving of gateway messages, and maintains the gateway connection and dispatch service. It expects the implementor to provide logic for connection and heartbeat sending, and provides an opportunity for implementors to process payloads as they are received.

### The `DiscordGatewayClient`

This has been refactored to implement the `BaseGatewayClient`. This has largely been a copy-paste operation; the main changes can be found in the heart-beating logic and the offhanding of send/receive to the base.

### `IResponderDispatchService` and `DefaultResponderDispatchService`

These are rather similar to the `ResponderDispatchService` that was recently introduced. They represent a means to dispatch payloads to relevant responders, and to handle the results of any responder execution.

To compare to the current implementation, my dispatch service uses a blocking run method, rather than start and stop. Hence, the consumer must manage the run task. I believe this has the advantage of the consumer knowing if a fatal error has occurred within the dispatch service, and can restart it correspondingly.

Further, this implementation expects `IGatewayEvent` instances to be enqueued, rather than `IPayload` instances.

Finally, I use an unbounded channel to enqueue payloads. My logic here being that I don't want to block the gateway receiver. Would very much like to hear more thoughts on this though; I can see that being unable to dispatch events fast enough is as good as having a blocked receiver.

### The `WebSocketPayloadTransportService`

I have made significant changes to this service. Primarily, I was aiming to reduce memory consumption. Relevant benchmarks are attached below.

I've also adjusted how new `ClientWebSocket` instances are resolved, and how the socket receive task is cancelled. These changes prevent the socket from being prematurely disposed or aborted upon any cancellation, allowing a proper close handshake to be completed.

[Payload Deserialization Benchmarks](https://github.com/carlst99/Remora.Discord/blob/m_voice/Tools/Remora.Discord.Benchmarks/PayloadDeserializationBenchmarks.cs).
```
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
AMD Ryzen 5 3600, 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.201
  [Host]     : .NET 6.0.3 (6.0.322.12309), X64 RyuJIT
  DefaultJob : .NET 6.0.3 (6.0.322.12309), X64 RyuJIT


|                                       Method |     Mean |    Error |   StdDev |   Median | Ratio | RatioSD |  Gen 0 |  Gen 1 | Allocated |
|--------------------------------------------- |---------:|---------:|---------:|---------:|------:|--------:|-------:|-------:|----------:|
|                           CurrentDeserialize | 91.39 us | 1.817 us | 4.592 us | 89.84 us |  1.00 |    0.00 | 7.3242 | 1.4648 |     60 KB |
| CurrentDeserializeWithRecyclableMemoryStream | 77.76 us | 0.596 us | 0.557 us | 77.58 us |  0.81 |    0.04 | 5.8594 | 1.0986 |     48 KB |
```

[Payload Serialization Benchmarks](https://github.com/carlst99/Remora.Discord/blob/m_voice/Tools/Remora.Discord.Benchmarks/PayloadSerializationBenchmarks.cs).
```
|              Method |     Mean |    Error |   StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|-------------------- |---------:|---------:|---------:|------:|--------:|-------:|----------:|
|    CurrentSerialize | 11.91 us | 0.228 us | 0.244 us |  1.00 |    0.00 | 0.4272 |   3,592 B |
|        NewSerialize | 10.76 us | 0.154 us | 0.137 us |  0.91 |    0.02 | 0.0458 |     424 B |

```

### Next Steps

More thorough unit testing might be needed; for example to ensure that pre-shutdown commands are properly sent.
I will also be testing this revised gateway in my bots over a number of days, to see if anything trips it up.


Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
```diff
@@ -0,0 +1,572 @@
...
+public abstract class BaseGatewayClient : IGatewayClient, IAsyncDisposable
+{
+    private readonly ILogger<BaseGatewayClient> _logger;
+    private readonly Channel<IPayload> _payloadSendQueue;
+    private readonly Dictionary<Guid, IGatewayCommand> _preShutdownCommands;
+
+    private Task<Result>? _sendTask;
+    private Task<Result>? _receiveTask;
+    private Task<Result>? _responderDispatchTask;
+    private volatile bool _isRunning;
```
VelvetThePanda, line 54: I think it may be better to use a normal bool here and use `Volatile.Read`/`Volatile.Write` instead

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
```diff
@@ -0,0 +1,572 @@
...
+
+            while (!ct.IsCancellationRequested)
+            {
+                switch (this.Status)
+                {
+                    case GatewayConnectionStatus.Offline:
+                    case GatewayConnectionStatus.Disconnected:
+                    {
+                        // We might have switched state while running. This is a signal to stop and reconnect.
+                        if (_isRunning)
```
VelvetThePanda, line 126: re: `Volatile`, especially considering the field is being used directly

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
```diff
@@ -0,0 +1,572 @@
...
+                        _isRunning = true;
+                        this.DisconnectRequestedSource.Dispose();
+                        this.DisconnectRequestedSource = new CancellationTokenSource();
+
+                        var connectResult = await ConnectAsync(ct);
+                        if (!connectResult.IsSuccess)
+                        {
+                            _logger.LogWarning
+                            (
+                                "Failed to connect to the gateway. Connection will be reattempted\n{Error}",
```
VelvetThePanda, line 144: I think *may* is better grammar here, since a failure may lead to the gateway stopping entirely.

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
```diff
@@ -0,0 +1,572 @@
...
+            return ex;
+        }
+    }
+
+    /// <summary>
+    /// Checks if a result should allow the client to reconnect, then stops the client.
+    /// </summary>
+    /// <param name="operationResult">The result.</param>
+    /// <returns>A value indicating whether or not to attempt a reconnection.</returns>
+    private async Task<bool> StopAndCheckResultForReconnection(Result operationResult)
```
VelvetThePanda, line 555: Missing `Async` postfix

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
```diff
@@ -0,0 +1,572 @@
...
+    /// </returns>
+    protected abstract Task<Result<TimeSpan>> SendHeartbeatAsync(CancellationToken ct);
+
+    /// <summary>
+    /// Connects the websocket and begins the send task.
+    /// </summary>
+    /// <param name="gatewayUri">The URI of the voice gateway.</param>
+    /// <param name="ct">A <see cref="CancellationToken"/> that can be used to stop the operation.</param>
+    /// <returns>A <see cref="Result"/> indicating the outcome of the operation.</returns>
+    protected async Task<Result> ConnectToGatewayAndBeginSendTask
```
VelvetThePanda, line 321: Missing `Async` postfix

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -423,225 +348,113 @@ private bool ShouldReconnect
...
-                break;
-            }
+    /// <inheritdoc />
+    protected override async Task<Result<TimeSpan>> SendHeartbeatAsync(CancellationToken ct)
+    {
+        if (_heartbeatData.Interval == TimeSpan.Zero)
+        {
+            // We have not received OP code hello yet, so don't heartbeat
+            return TimeSpan.FromMilliseconds(50);
         }
```
VelvetThePanda, line 415: More of a nitpick/nuance thing but [according to the docs](https://discord.dev/topics/gateway#heartbeating), you could *technically* heartbeat immediately. Not that you should, but you *could*.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -677,23 +486,23 @@ private async Task<Result> CreateNewSessionAsync(CancellationToken ct = default)
...
-                shardInformation,
-                initialPresence,
-                _gatewayOptions.Intents
-            )
+                Intents: _gatewayOptions.Intents,
+                Compress: false,
+                Shard: shardInformation,
+                Presence: initialPresence
+            ),
+            ct
```
VelvetThePanda, line 500: `LargeThreshold` appears to be absent. Is your branch out of date?

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -706,17 +515,12 @@ private async Task<Result> CreateNewSessionAsync(CancellationToken ct = default)
 
             if (receiveReady.Entity is IPayload<IInvalidSession> invalidSession)
             {
-                return new GatewayError
-                (
-                    "The newly created session was invalidated by Discord.",
-                    invalidSession.Data.IsResumable,
-                    false
-                );
+                return new GatewayError("Session invalidated", false, false);
```
VelvetThePanda, line 518: Why does this no longer attempt to resume?

Backend/Remora.Discord.Gateway/IGatewayClient.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// </summary>
+    TimeSpan Latency { get; }
+
+    /// <summary>
+    /// Runs the gateway client. This task will only complete when the
+    /// gateway connection fatally disconnects, or it is cancelled.
+    /// </summary>
+    /// <param name="ct">A <see cref="CancellationToken"/> that can be used to stop the operation.</param>
+    /// <returns>A result representing the outcome of the operation.</returns>
+    Task<Result> RunAsync(CancellationToken ct = default);
```
VelvetThePanda, line 54: There's a useful feature in which you can resume at startup, but given that the GW client implements this interface, I don't think it's possible to do this is it?

Could always make it an overload, but eh.

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
```diff
@@ -55,27 +68,41 @@ public class WebSocketPayloadTransportService : IPayloadTransportService, IAsync
...
-        IServiceProvider services,
+        RecyclableMemoryStreamManager memoryStreamManager,
         JsonSerializerOptions jsonOptions
     )
     {
-        _services = services;
+        _memoryStreamManager = memoryStreamManager;
         _jsonOptions = jsonOptions;
+
+        _sendSemaphore = new SemaphoreSlim(1, 1);
```
VelvetThePanda, line 82: nitpick: Rider should be giving a hint that `new SemaphoreSlim(1)` is equivalent

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
```diff
@@ -113,56 +140,53 @@ public async Task<Result> ConnectAsync(Uri endpoint, CancellationToken ct = defa
...
 
-        await using var memoryStream = new MemoryStream();
-
-        byte[]? buffer = null;
         try
         {
-            await JsonSerializer.SerializeAsync(memoryStream, payload, _jsonOptions, ct);
-
-            if (memoryStream.Length > 4096)
+            semaphoreEntered = await _sendSemaphore.WaitAsync(100, ct);
```
VelvetThePanda, line 152: Interesting use of WaitAsync.

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
```diff
@@ -172,58 +196,58 @@ public async Task<Result> SendPayloadAsync(IPayload payload, CancellationToken c
...
         try
         {
-            WebSocketReceiveResult result;
+            semaphoreEntered = await _receiveSemaphore.WaitAsync(100, ct);
+            if (!semaphoreEntered)
+            {
+                return new OperationCanceledException("Could not enter semaphore.");
+            }
 
+            await using var ms = _memoryStreamManager.GetStream();
```
VelvetThePanda, line 214: I think Jax would advise against abbreviated variable names like this 😄

VelvetThePanda (changes requested):
Chonky PR! Good work, though. I've left some comments and reviews. Looks good outside of what's explicitly pointed out here.

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
```diff
@@ -0,0 +1,572 @@
...
+
+/// <summary>
+/// Provides an abstract implementation of an <see cref="IGatewayClient"/>
+/// that must be extended with relevant connection and resumption flows.
+/// </summary>
+public abstract class BaseGatewayClient : IGatewayClient, IAsyncDisposable
+{
+    private readonly ILogger<BaseGatewayClient> _logger;
+    private readonly Channel<IPayload> _payloadSendQueue;
+    private readonly Dictionary<Guid, IGatewayCommand> _preShutdownCommands;
```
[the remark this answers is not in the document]
VelvetThePanda: Looking back, I can see why; a GUID is an...interesting choice, however.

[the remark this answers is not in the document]
carlst99: I'm open to suggestions on the use of a GUID. I was simply looking for an identifier that I could always guarantee to be unique.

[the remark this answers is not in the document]
VelvetThePanda: Well in terms of purely BCL things I think that's about as good as it gets.

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
carlst99: Velvet out here with their fishing rod 😂. Another good catch, cheers

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
carlst99: I made that change because the docs say this lol:

> You may send heartbeats before this interval elapses, but you should avoid doing so unless necessary

So yeah, we could. But, likewise, I don't think we should 😛.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
carlst99: Yes, it was. I did attempt to do a merge but it was 0330 at that point. Thanks, I'll have to comb over again.

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
VelvetThePanda: Aye. I was thinking of a video I saw about the wonky behavior of `volatile` but looking at what Fred has said about it I don't think it matters here. 
D3542525-6037-440F-BC55-F35227604D8B

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
carlst99: Using the `IsResumable` property appeared to result in a developer error... fixed 😄

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
carlst99: In what way? With the current use of the transport service, the send method will never be called from more than two threads. My line of thinking is hence, if it has taken that long to enter the semaphore, something has probably gone wrong. This is certainly expecting things to be a bit more domain specific than the accessibility of the service might otherwise indicate.

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
carlst99: You're probably right... given the existing name was `memoryStream`, I'll change it back lol

Backend/Remora.Discord.Gateway/IGatewayClient.cs
carlst99: I'll look into this

carlst99:
Testing in my bot over a longer period has shown there is a significant issue somewhere; the bot keeps failing to make interaction responses in time. This likely means that either the receive loop or dispatch logic has a bottleneck somewhere. Regardless, this isn't ready to see the light of day.

Backend/Remora.Discord.Gateway/Services/DefaultResponderDispatchService.cs
```diff
@@ -0,0 +1,391 @@
...
+                {
+                    var dispatchResult = DispatchEvent(gatewayEvent);
+                    if (!dispatchResult.IsSuccess)
+                    {
+                        return Result.FromError(dispatchResult);
+                    }
+
+                    await _responderFinalizationQueue.Writer.WriteAsync(dispatchResult.Entity, ct);
+
+                    if (_finalizerTask.IsCompleted)
```
VelvetThePanda, line 130: Is this meant to be inverted? `awaiting` a completed task is essentially a no-op, and given that it's completed, you could return `.Result`, though at a glance you may get a few looks for that.

carlst99: I agree; it should be inverted so that this check is made before the rest of the logic. At least, I assume that's what you meant?

As to `awaiting` it, am I mistaken in thinking that it will resolve to returning the result directly? I see no harm in changing it though. I've kept consistent with how Jax has written similar logic - see [here](https://github.com/Nihlus/Remora.Discord/blob/feaf50c481cbdbdab025635aa73804a7f7a4ed8f/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs#L591).

VelvetThePanda: Hmm, looks like Github selected more than intended; I was referring to the boolean check itself, hence the comment on awaiting a completed task. Consistency is good, though.

carlst99: Ah, gotcha. No, the finalizer task should run for the lifetime of the dispatch service. If it has completed then something has gone wrong, so I return that error. It is then up to the consumer to inspect the error and restart the dispatch service.

Nihlus:
Awesome work! I definitely like the blocking Run method better than the Start/Stop pattern I currently use. 

Unfortunately, I'm swamped with a lot of other stuff at the moment, so I won't be able to do a proper review for a while - that said, it sounds like you might need some time to get things into a stable state.

Drop a comment here when you're reading for a thorough review, but overall, I like the design.

carlst99:
Thanks! No problem at all, I'll leave an update when I think it's ready to go. I certainly need some time to look into a couple of issues, and that may take a little while myself.

Foxtrek64:
I wonder if some of this should be pulled and sent up to Remora.Rest or Remora.Rest.Gateway. I could see the this being useful for more than just Discord.

AraHaan:
Wait why do I see await calls not call ``.ConfigureAwait(false);`` here to avoid the chance have deadlocking every thread in the any program that use remora?

Nihlus:
...because that's not what ConfigureAwait(false) is for?

carlst99:
> Testing in my bot over a longer period has shown there is a significant issue somewhere; the bot keeps failing to make interaction responses in time...

Further testing has proven the implementation to appear far more stable now. I'm less sure that it's to do with any changes I've made, rather than that my test bed had a couple of glaring flaws 😬.
I've also run the SlashCommands sample through the dotTrace profiler, and can't see anything that looks out of the ordinary. I'm not very familiar with profiling, mind.

I'd still like to do another pass over the code before a full-on review, but I think it's closer to a ready state now.

VelvetThePanda:
It seems you dropped support for OCE when logging errors?

carlst99:
Yes, either an `OperationCanceledException` or `TaskCanceledException` should indicate a user-requested shutdown. I didn't feel a need to log them, but it's no harm to do so.

VelvetThePanda:
Ah. Sorry, missed line 109 somehow

carlst99 (commented):
Would like to point out this potentially breaking change, in that payloads with an unrecognised OP code will be deserialized to an `IUnknownEvent`, rather than throwing an `ArgumentOutOfRangeException`.

Edit: It didn't link -_-. See here https://github.com/Nihlus/Remora.Discord/pull/176/commits/00409d559db9496e8604145bb0690c19a90a8454

Backend/Remora.Discord.Gateway/IGatewayClient.cs
carlst99: Done!

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -366,21 +407,27 @@ protected override async ValueTask<bool> ProcessPayloadAsync(IPayload payload, C
             UpdateLastReceivedHeartbeatToNow();
         }
 
-        // Signal the governor task that a reconnection is requested, if necessary.
         switch (payload)
         {
             case IPayload<IReconnect>:
             {
+                _logger.LogDebug("Reconnect requested");
```
VelvetThePanda, line 414: Postfix with ` by gateway`?

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -366,21 +407,27 @@ protected override async ValueTask<bool> ProcessPayloadAsync(IPayload payload, C
...
                 this.Status = GatewayConnectionStatus.Disconnected;
 
                 break;
             }
             case IPayload<IInvalidSession> invalidSession:
             {
+                _logger.LogDebug("Session invalidated");
```
VelvetThePanda, line 421: Ditto

VelvetThePanda (changes requested):

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
```diff
@@ -454,7 +455,7 @@ private async Task<Result> GatewayReceiverAsync(CancellationToken ct)
 
             while (!ct.IsCancellationRequested)
             {
-                var receiveResult = await this.TransportService.ReceivePayloadAsync(ct);
+                var receiveResult = await this.TransportService.ReceivePayloadAsync(ct).ConfigureAwait(false);
```
VelvetThePanda, line 456: Any reason for the `CA(false)` here?

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
carlst99: Sounds good

Backend/Remora.Discord.Gateway/BaseGatewayClient.cs
carlst99: This is a call that will be waited on an awful lot, so I've followed the guidelines laid out [here](https://devblogs.microsoft.com/dotnet/configureawait-faq/) - this is contained code that should have no need to be invoked on a particular scheduler and I'd like to avoid the associated overhead.

VelvetThePanda:
> Cant we have this be instead of 50 to be lower to like 30 or 20?
Why would we want to burn more CPU cycles? 50ms is fine imho.

AraHaan:
Can this be worked on more? I would most certainly like this.

carlst99:
> Can this be worked on more? I would most certainly like this.

Sorry for the slow reply. I've had a number of other projects on my plate. With the existing gateway having moved ahead and not foreseeing any immediate time to restart work on the voice implementation, I (likely mistakenly) let this fall by the wayside.

I'll likely start over and rebase relevant changes onto the current gateway implementation. I'll scope this out over the next week and update then if I think I'll have it done in any sort of reasonable timeframe. Otherwise, probably best to drop it so other people feel more free to PR their own implementation.

AraHaan:
That is ok, as long as we can eventually get a common gateway that would be ok with me.

I would also be ok with making the common gateway, but split it into it's own library that is not specific to discord as well.

Foxtrek64:
Having a generic gateway client in its own library (maybe `Remora.Rest.Gateway`?) could be rather useful. I can personally think of a couple places where I'd have a use for it, so I'm sure I wouldn't be alone here.

AraHaan:
I could see a need for it for cases like:

- you want to make a custom wrapper to an api with a gateway similar to Discord's.

For that case a remora generic gateway library would be a great idea and also allows more code to be shared as well (that is not only Discord specific gateway stuff).

carlst99:
I don't think this is suitable to break out into a generic library. Most of the components have logic that's tailored towards Discord's error codes and connection specifics. The event dispatch system is free of this, but I'd argue it's fairly easy to replicate, and I imagine many people will be using their own dispatch mechanisms regardless.

Anecdotally, I maintain a library which also wraps an API with a websocket event stream. I'd not be able to reuse the gateway client, and would be making a number of changes to the transport service. My dispatch system is similar, but has a few critical differences.
To write Remora's gateway logic in such a way that it's generic would require an impractically high level of abstraction, I feel.

carlst99:
Well, we're another two months down the line and I've still made no progress on this 😓. I still hope to revisit this at some stage, but in the meantime I'm going to close this PR, particularly given how out of date it currently is.

PR #179, opened 2022-04-06
Parse X-RateLimit-Reset independent of system culture

MazeXP:
The `X-RateLimit-Reset` header is currently parsed based on the systems locale.

Some locales use a `,` as decimal separator and some locales use `.` for separation.

Therefore the double must be parsed indepedent of the systems locale.

### DIFF
--- a/Backend/Remora.Discord.Rest/API/RateLimitBucket.cs
+++ b/Backend/Remora.Discord.Rest/API/RateLimitBucket.cs
@@ -22,6 +22,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Linq;
 using System.Net.Http.Headers;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


Backend/Remora.Discord.Rest/API/RateLimitBucket.cs
```diff
@@ -112,7 +113,7 @@ public static bool TryParse(HttpResponseHeaders headers, [NotNullWhen(true)] out
                 return false;
             }
 
-            if (!double.TryParse(rawReset.SingleOrDefault(), out var resetsAtEpoch))
+            if (!double.TryParse(rawReset.SingleOrDefault(), NumberStyles.Float | NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture, out var resetsAtEpoch))
```
Nihlus, line 116: Use `NumberStyles.Float | NumberStyles.AllowThousands` instead, as that's the default flag set.

Nihlus (changes requested):
Thanks! Just one small change and this is good to go :)

PR #185, opened 2022-04-27
Use delegates instead of MethodInfo for dispatch

VelvetThePanda:
This PR adds an optimization for responder dispatch which replaces `MethodInfo` calls with delegates. In theory dispatch is now 17 times faster, but actual performance gains will probably be negligible. Below are the results of benchmarks of purely just MethodInfo versus delegate calls:

```fix
|          Method |       Mean |     Error |    StdDev | Allocated |
|---------------- |-----------:|----------:|----------:|----------:|
|      MethodInfo | 123.626 ns | 1.8388 ns | 1.5355 ns |      32 B |
| UnboundDelegate |   8.015 ns | 0.1286 ns | 0.1203 ns |         - |
|   BoundDelegate |   7.197 ns | 0.0738 ns | 0.0691 ns |         - |
```

### DIFF
--- a/Backend/Remora.Discord.Gateway/Services/ResponderDispatchService.cs
+++ b/Backend/Remora.Discord.Gateway/Services/ResponderDispatchService.cs
@@ -48,7 +48,7 @@ public class ResponderDispatchService : IAsyncDisposable
     private readonly IResponderTypeRepository _responderTypeRepository;
 
     private readonly Dictionary<Type, Type> _cachedInterfaceTypeArguments;
-    private readonly Dictionary<Type, MethodInfo> _cachedDispatchMethods;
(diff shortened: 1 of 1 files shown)


VelvetThePanda:
This PR is based off the findings of [VelvetThePanda/DelegatePOC](https://github.com/VelvetThePanda/DelegatePOC), and a conversation on the Discord about whether we could get away with using delegates instead of method info. After some trial and error, the answer is yes!

AraHaan:
Is this PR ready?

VelvetThePanda:
It'd be marked as a draft otherwise, yes.

AraHaan:
Are tests needed for this as well too?

VelvetThePanda:
I've validated this concept works in [my own repo](https://github.com/VelvetThePanda/DelegatePOC).

PR #149, opened 2022-01-18
Slash-Command V2 Permissions support

VelvetThePanda:
It is what it says on the tin; I'll continue to update this as changes happen/as required

PR #187, opened 2022-04-28
Filter out null elements from attribute list

teraa:
`GetCustomAttribute` returns null if no matching attribute is found. This fixes the following `Count` check being incorrect.


AraHaan:
Due to the permissions v2 changes breaking the lib on this, I think the PR can be closed as there is currently another PR implementing permissions v2 (and the removal of the attribute the PR is trying to correct).

Nihlus:
Superseded by #149.

PR #194, opened 2022-05-11
Implement interaction response tracking

Nihlus:
This PR implements stateful tracking of whether an interaction has been responded to via the interaction context. This lets the feedback service (among other things) respond appropriately when a contextual message is sent instead of failing outright.

However, this has required some rethinking regarding overriding service registrations (and one ugly hack), which means that a significant overhaul of the caching REST implementations is also part of this PR. Namely, instead of using inheritance and overrides, the services now use the decorator pattern, forwarding calls to a field instead.

This has several benefits, but first and foremost it allows more flexible customization of the REST API types, removing the need to rely on an inheritance chain if behaviour changes or additions are required - any package can now step into the execution path of REST calls and do whatever it needs without interfering with any other implementations that do the same.

It is, however, a rather significant shift in design for these types, so I'd very much appreciate feedback on this PR before I merge it.

EDIT 1:

Based on user feedback, I've also rolled in a reworking of how REST customizations are used, since these changes would make customizations far more unreliable (and incompatible with caching). In short, instead of injecting the concrete type and performing customizations on that, users should now check for the presence of the `IRestCustomizable` interface on the injected interface and operate on that.

```cs
IDiscordRest...API _api;
if (_api is IRestCustomizable customizable)
{
    using var customization = customizable.WithCustomization(b => ...);
}
```

This comes with a caveat, however - if the injected interface does *not* implement `IRestCustomizable`, the decorators will *throw* when you try to add a customization. If you're using Remora's default types, you'll be fine since they all implement this interface, but you may run into issues with external implementations (if any).

### DIFF
--- a/Remora.Discord.Commands/API/ResponseTrackingInteractionAPI.cs
+++ b/Remora.Discord.Commands/API/ResponseTrackingInteractionAPI.cs
@@ -1,0 +1,212 @@
+//
+//  ResponseTrackingInteractionAPI.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) 2017 Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using OneOf;
+using Remora.Discord.API.Abstractions.Objects;
+using Remora.Discord.API.Abstractions.Rest;
+using Remora.Discord.Commands.Contexts;
+using Remora.Discord.Commands.Services;
+using Remora.Rest.Core;
+using Remora.Results;
+
+namespace Remora.Discord.Commands.API;
+
+/// <summary>
+/// Implements a tracking system for whether an interaction has been responded to.
+/// </summary>
+internal class ResponseTrackingInteractionAPI : IDiscordRestInteractionAPI
+{
+    private readonly ContextInjectionService? _contextInjector;
+    private readonly IDiscordRestInteractionAPI _actual;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ResponseTrackingInteractionAPI"/> class.
+    /// </summary>
+    /// <param name="actual">The actual API instance.</param>
+    /// <param name="provider">The service provider.</param>
+    public ResponseTrackingInteractionAPI
+    (
+        IDiscordRestInteractionAPI actual,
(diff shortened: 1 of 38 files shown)


Remora.Discord.Commands/API/ResponseTrackingInteractionAPI.cs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// <inheritdoc />
+    public Task<Result<IMessage>> GetOriginalInteractionResponseAsync
+    (
+        Snowflake applicationID,
+        string interactionToken,
+        CancellationToken ct = default
+    ) =>
+        _actual.GetOriginalInteractionResponseAsync(applicationID, interactionToken, ct);
```
VelvetThePanda, line 104: Couldn't an optimization technically be made here to immediately return a `RestResult<NotFoundError>` if a response hasn't been created yet? Would save on API calls.

Remora.Discord.Commands/API/ResponseTrackingInteractionAPI.cs
```diff
@@ -0,0 +1,212 @@
...
+        );
+
+    /// <inheritdoc />
+    public Task<Result> DeleteOriginalInteractionResponseAsync
+    (
+        Snowflake applicationID,
+        string token,
+        CancellationToken ct = default
+    ) =>
+        _actual.DeleteOriginalInteractionResponseAsync(applicationID, token, ct);
```
VelvetThePanda, line 137: ditto

VelvetThePanda (approved):
👏🏽 Looks pretty decent from what I can see; I'll definitely need to read up more on the decorator pattern, maybe talk it out with Nick Chapsas, seeing as he's gone over this before.

Remora.Discord.Commands/API/ResponseTrackingInteractionAPI.cs
MazeXP: It could not be optimized by simply returning RestResult<NotFoundError> without breaking every other class that is using `IDiscordRestInteractionAPI` as in the `ServiceCollection` the implementation of `IDiscordRestInteractionAPI` is being replaced by `ResponseTrackingInteractionAPI` while wrapping the old registered `IDiscordRestInteractionAPI`.

The calling chain due to the decorator pattern would be as follows: (caching enabled + commands enabled)
```
ResponseTrackingInteractionAPI.GetOriginalInteractionResponseAsync(...) 
calls (delegates to)
  CachingDiscordRestInteractionAPI.GetOriginalInteractionResponseAsync(...) 
  calls (delegates to)
    DiscordRestInteractionAPI .GetOriginalInteractionResponseAsync(...)
```

So any class that is using `IDiscordRestInteractionAPI` would break when you break the delegation chain (by not calling the actual method).

MazeXP (approved):
LGTM

Remora.Discord.Commands/API/ResponseTrackingInteractionAPI.cs
```diff
@@ -0,0 +1,212 @@
...
+    /// <param name="provider">The service provider.</param>
+    public ResponseTrackingInteractionAPI
+    (
+        IDiscordRestInteractionAPI actual,
+        IServiceProvider provider
+    )
+    {
+        _actual = actual;
+
+        // TODO: Improve this with a proper scope check once I figure out how to do that
```
MazeXP, line 59: Should it not just work by having it as constructor parameter?
As the scoped ContextInjectionService will be nevertheless will be injected each time in the transient IDiscordRestInteractionAPI.
Or do you actually want to detect if the REST API is being instantiated inside of a scope?

Nihlus: Unfortunately, using a straight constructor parameter doesn't work when resolving the API instance from the root service provider, which is still a use case I want to support. If there's no context in play, the service should still be usable as-is.

PR #200, opened 2022-06-15
Add options to configure the DiscordService to check slash command support and optionally update slash commands.

AraHaan:
Fixes https://github.com/Nihlus/Remora.Discord/issues/199.


Remora.Discord.Hosting/Services/DiscordService.cs
```diff
@@ -69,6 +75,53 @@ ILogger<DiscordService> logger
...
+            _logger.LogInformation("Checking for slash commands support.");
+            var checkSlashSupport = _slashService.SupportsSlashCommands();
+            if (!checkSlashSupport.IsSuccess)
+            {
+                // write to log.
+                _logger.LogError(
+                    "The registered commands of the bot don't support slash commands: {Message}",
+                    checkSlashSupport.Error.Message);
+
+                // Write to Console as well in case log is disabled or filtered to not show in console.
```
VelvetThePanda, line 90: Never ever do this.

Remora.Discord.Hosting/Services/DiscordService.cs
```diff
@@ -69,6 +75,53 @@ ILogger<DiscordService> logger
     /// <inheritdoc />
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
+        // do not check after the initial check at bot startup.
```
VelvetThePanda, line 78: This method is called once, so I find this comment to be moot?

VelvetThePanda: Well, once*, but even then changing host background service faulting behavior is niche in it of itself.

Remora.Discord.Hosting/Services/DiscordService.cs
```diff
@@ -69,6 +75,53 @@ ILogger<DiscordService> logger
...
+                    var updateSlash = await _slashService.UpdateSlashCommandsAsync(
+                        _options.UpdateSlashGuild, ct: default).ConfigureAwait(false);
+                    if (!updateSlash.IsSuccess)
+                    {
+                        // write to log.
+                        _logger.LogError(
+                            "Failed to update slash commands: {0}",
+                            updateSlash.Error.Message);
+
+                        // Write to Console as well in case log is disabled or filtered to not show in console.
```
VelvetThePanda, line 112: See above. Repugnant behavior from a library.

Remora.Discord/Remora.Discord.csproj
```diff
@@ -5,18 +5,12 @@
...
-            Update voice channel endpoints and permissions.
-            Remove audit log reason from modify channel position.
-            Add new error code.
-            Implement Modify Guild MFA Level.
-            Add new error code.
-            Break out event-only fields.
-            Add new error codes.
-        </PackageReleaseNotes>
+			Add options to configure the DiscordService to check slash command support and optionally update slash commands.
+			Update dependencies.
```
VelvetThePanda, line 12: Redundant/misleading update message.

Remora.Discord.Hosting/Options/DiscordServiceOptions.cs
```diff
@@ -28,4 +30,18 @@ namespace Remora.Discord.Hosting.Options;
...
+/// </param>
+/// <param name="UpdateSlashGuild">
+/// The guild that the service should update slash commands for.
+/// When null, it updates global commands.
+/// </param>
+public record DiscordServiceOptions(
+    bool TerminateApplicationOnCriticalGatewayErrors = true,
+    bool CheckSlashCommandsSupport = false,
+    bool UpdateSlashCommands = false,
+    Snowflake? UpdateSlashGuild = null);
```
VelvetThePanda, line 47: Allman parentheses please.

Remora.Discord.Hosting/Remora.Discord.Hosting.csproj
```diff
@@ -5,15 +5,16 @@
     </PropertyGroup>
 
     <PropertyGroup>
-        <Version>5.0.17</Version>
+        <Version>5.0.18</Version>
```
VelvetThePanda, line 8: I feel it's somewhat out of scope to update the version in a PR, but I digress.

Remora.Discord.Hosting/Remora.Discord.Hosting.csproj
```diff
@@ -5,15 +5,16 @@
...
-            Update dependencies.
-        </PackageReleaseNotes>
+			Add options to configure the DiscordService to check slash command support and optionally update slash commands.
+			Update dependencies.
+		</PackageReleaseNotes>
     </PropertyGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\Backend\Remora.Discord.Gateway\Remora.Discord.Gateway.csproj" />
+        <ProjectReference Include="..\Remora.Discord.Commands\Remora.Discord.Commands.csproj" />
```
VelvetThePanda, line 17: Is there a reason this was *replaced*? I question whether this code even compiles.

VelvetThePanda (changes requested):

Remora.Discord.Hosting/Remora.Discord.Hosting.csproj
AraHaan: Remora.Discord.Commands includes Gateway and I seen that other remora libs do not reference directly a dependency of a dependency manually so I changed this line only to be Remora.Discord.Commands instead.

Remora.Discord.Hosting/Services/DiscordService.cs
AraHaan: I thought it could be called multiple times in the case of direct “return;" to loop back and recall it.

VelvetThePanda: That would both be non-trivial to determine, and rather asinine behavior. [See the docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.backgroundservice.executeasync)

Remora.Discord.Hosting/Options/DiscordServiceOptions.cs
```diff
@@ -1,31 +1,49 @@
-//
```
AraHaan, line 1: 😱 Visual Studio, why?

MazeXP:
I do not think that option would be useful except the Hosting package should enforce a dependency on Remora.Discord.Commands.
Even for developers that do not use Commands in their bots at all.

MazeXP:
I would prefer some extensions method on the `Host` class itself.

*Usage*
```cs
Host.CreateDefaultBuilder(args)
    ...
    .Build()
    .SetupSlashCommands() // Choose a better name for the command
    .Start()
```

Hamsterland:
It might just need to be added as extra optional parameteres here: https://github.com/Nihlus/Remora.Discord/blob/master/Remora.Discord.Commands/Extensions/ServiceCollectionExtensions.cs#L120= 

All it would do then is add a separate `SlashCommandUpdaterHostedService` (change name) that adds and updates slash commands.

I personally don't like the idea of an extension method on the Host because of the async calls. But my idea is just another way to approach the problem. In the end, I think this is something the developer can do themselves.

Remora.Discord.Hosting/Remora.Discord.Hosting.csproj
AraHaan: For them, they could easily implement their own BackgroundService and use that for their own command system. For those who use Remora.Discord.Commands I say an option to add / update commands as well as check for support for slash is a good option in the default DiscordService inside Remora.Discord.Hosting (to make things simple for simple bots).

Nihlus:
> I do not think that option would be useful except the Hosting package should enforce a dependency on Remora.Discord.Commands. Even for developers that do not use Commands in their bots at all.

This is a very good point, and one I agree with - using a generic host should not also require you to use commands. Furthermore, this implementation complicates the use of multiple named trees, and is rather inflexible in its design.

As it stands, I don't see this PR getting merged - thanks for the suggestion, though!

PR #206, opened 2022-06-18
Add customization and parallelization to responder dispatching

VelvetThePanda:
This PR adds the ability to cuztomize and parallelize dispatch of responders. This is certainly tuned toward larger bots, but could bring some performance benefits overall.

It's worth noting that in very high throughput situations, parallelizing dispatch can cause race conditions in entity states, especially if using a backing store such as Redis.

### DIFF
--- a/Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
+++ b/Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
@@ -49,14 +49,12 @@ public static class ServiceCollectionExtensions
     /// <param name="serviceCollection">The service collection.</param>
     /// <param name="tokenFactory">A function that retrieves the bot token.</param>
     /// <param name="buildClient">Extra options to configure the rest client.</param>
-    /// <param name="maxDispatchItems">How many items can be concurrently queued for dispatch.</param>
     /// <returns>The service collection, with the services added.</returns>
     public static IServiceCollection AddDiscordGateway
     (
         this IServiceCollection serviceCollection,
         Func<IServiceProvider, string> tokenFactory,
-        Action<IHttpClientBuilder>? buildClient = null,
-        int maxDispatchItems = 100
+        Action<IHttpClientBuilder>? buildClient = null
     )
     {
         serviceCollection
@@ -75,7 +73,7 @@ public static class ServiceCollectionExtensions
             s.GetRequiredService<ILogger<WebSocketPayloadTransportService>>()
         ));
 
-        serviceCollection.Configure<ResponderDispatchOptions>(dispatch => dispatch with { MaxItems = (uint)maxDispatchItems });
+        serviceCollection.Configure<ResponderDispatchOptions>(() => new());
 
         return serviceCollection;
     }
--- a/Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
+++ b/Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
@@ -33,6 +33,7 @@ using Remora.Discord.Gateway.Responders;
 using Remora.Discord.Gateway.Services;
 using Remora.Discord.Gateway.Transport;
 using Remora.Discord.Rest.Extensions;
+using Remora.Extensions.Options.Immutable;
 
 namespace Remora.Discord.Gateway.Extensions;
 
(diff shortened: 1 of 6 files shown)


AraHaan:
Does it also include such issues for MemoryCache as well?

Nihlus:
What's the difference here? Dispatched events are not awaited by the dispatcher, and it will push them through as fast as it can process them. 

I would very much like to see some measurements for this PR.

VelvetThePanda:
Dispatch itself isn't the issue; the issue is *queuing* into dispatch. 

In retrospect this may be a micro-optimization, but the basic issue is:

Get event ➜ Queue to dispatch ➜ Wait for dispatch queue to not be full ➜ Loop back

In very high-throughput situations, it could be determined that there's a bottleneck in the gateway if you're receiving > 100 event/s, but I'm not sure how I'd even begin to measure that. 

This was initially opened because I assumed there was a bottleneck in Remora, but after talking to some large bot devs, and doing some additional testing, the issue is mostly with Discord.

Nihlus:
Personally, I'd prefer it if we kept the single dispatch task as it is, but merge the changes that allow you to configure the size of the queue - those are good options to expose to the end user.

VelvetThePanda:
Can do.

Backend/Remora.Discord.Gateway/ResponderDispatchOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
```
AraHaan, line 22: Add using to stop the warnings here?

MazeXP: Instead of a `using` only for the documentation `see` tag.
cref should point to the FullName with Namespace.

Nihlus: I'm fine with adding a `using` just for this - it's how it's done elsewhere in the library.

AraHaan:
```suggestion
using Remora.Discord.Gateway.Services;

```

Nihlus:
You've got failing tests; fix that and we're good to merge.

Tests/Remora.Discord.Gateway.Tests/Tests/DiscordGatewayClientTests.cs
```diff
@@ -101,6 +103,7 @@ public async Task CanConnectAsync()
         var transportMockDescriptor = ServiceDescriptor.Singleton(typeof(IPayloadTransportService), transportMock);
 
         var services = new ServiceCollection()
+            .Configure<ResponderDispatchOptions>(() => new())
```
Nihlus, line 106: You should do this in `AddDiscordGateway` instead; as it is, any consumer will also have to do this which is quite unintuitive.

Nihlus (changes requested):

Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -48,12 +49,14 @@ public static class ServiceCollectionExtensions
...
     /// <param name="buildClient">Extra options to configure the rest client.</param>
+    /// <param name="maxDispatchItems">How many items can be concurrently queued for dispatch.</param>
     /// <returns>The service collection, with the services added.</returns>
     public static IServiceCollection AddDiscordGateway
     (
         this IServiceCollection serviceCollection,
         Func<IServiceProvider, string> tokenFactory,
-        Action<IHttpClientBuilder>? buildClient = null
+        Action<IHttpClientBuilder>? buildClient = null,
+        int maxDispatchItems = 100
```
Nihlus, line 59: Don't take this as a parameter here - let the consumer use `Configure` if they want to change it.

Nihlus (changes requested):

Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
VelvetThePanda: Okay, I see what you mean.

PR #209, opened 2022-06-20
Attempt to eliminate annoying stack trace that gets logged when socket gets aborted.

AraHaan:
I think doing this instead of logging it's exception should be better than how it is currently. Aborted sockets cant be avoided, however it's stack trace on being logged I think can.

### DIFF
--- a/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
+++ b/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
@@ -1047,16 +1047,6 @@ public class DiscordGatewayClient : IDisposable
 
                 var receivedAt = DateTime.UtcNow;
 
-                if (receivedPayload.Error is Results.WebSocketError { State: WebSocketState.Aborted })
-                {
-                    // request reconnect.
-                    _shouldReconnect = true;
-
-                    // idk if connection is resumable.
-                    _isSessionResumable = false;
-                    return Result.FromSuccess();
-                }
-
                 if (!receivedPayload.IsSuccess)
                 {
                     // Normal closures are okay
--- a/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
+++ b/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
@@ -1047,6 +1047,16 @@ public class DiscordGatewayClient : IDisposable
 
                 var receivedAt = DateTime.UtcNow;
 
+                if (receivedPayload.Error is Results.WebSocketError { State: WebSocketState.Aborted })
+                {
+                    // request reconnect.
+                    _shouldReconnect = true;
+
+                    // idk if connection is resumable.
+                    _isSessionResumable = false;
+                    return Result.FromSuccess();
+                }
+
                 if (!receivedPayload.IsSuccess)
                 {
                     // Normal closures are okay
--- a/Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
+++ b/Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
@@ -247,8 +247,25 @@ public class WebSocketPayloadTransportService : IPayloadTransportService, IAsync
 
             do
             {
-                result = await _receiveRetryPolicy.ExecuteAsync(c => _clientWebSocket.ReceiveAsync(buffer, c), ct);
+                result = await _receiveRetryPolicy.ExecuteAsync
+                (
+                    async c =>
+                    {
+                        WebSocketReceiveResult result1;
+                        try
+                        {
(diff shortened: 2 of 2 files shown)


Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -1047,6 +1047,16 @@ private async Task<Result> GatewayReceiverAsync(TimeSpan heartbeatInterval, Canc
...
                 var receivedAt = DateTime.UtcNow;
 
+                if (receivedPayload.Error is Results.WebSocketError { State: WebSocketState.Aborted })
+                {
+                    // request reconnect.
+                    _shouldReconnect = true;
+
+                    // idk if connection is resumable.
+                    _isSessionResumable = false;
+                    return Result.FromSuccess();
```
AraHaan, line 1057: I am not completely sure if it should use ``FromSuccess()``, or simply use ``FromError(receivedPayload);`` here.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -1047,6 +1047,16 @@ private async Task<Result> GatewayReceiverAsync(TimeSpan heartbeatInterval, Canc
 
                 var receivedAt = DateTime.UtcNow;
 
+                if (receivedPayload.Error is Results.WebSocketError { State: WebSocketState.Aborted })
+                {
+                    // request reconnect.
+                    _shouldReconnect = true;
+
+                    // idk if connection is resumable.
+                    _isSessionResumable = false;
```
AraHaan, line 1056: I am not sure if it should be resumable so I set it to false in case to start a completely new session.

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
```diff
@@ -247,8 +247,25 @@ public async Task<Result<IPayload>> ReceivePayloadAsync(CancellationToken ct = d
...
+                    {
+                        WebSocketReceiveResult result1;
+                        try
+                        {
+                            result1 = await _clientWebSocket.ReceiveAsync(buffer, c);
+                        }
+                        catch
+                        {
+                            await DisconnectAsync(true, c);
+                            throw;
```
VelvetThePanda, line 262: Disconnecting and re-throwing is not what should be done here; re-init the socket and attempt to connect again.

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
```diff
@@ -276,6 +293,11 @@ public async Task<Result<IPayload>> ReceivePayloadAsync(CancellationToken ct = d
 
             return Result<IPayload>.FromSuccess(payload);
         }
+        catch
+        {
+            // this is most likely reached when the WebSocketState is aborted.
+            return new Results.WebSocketError(WebSocketState.Aborted);
+        }
```
VelvetThePanda, line 300: See above; furthermore don't make guesses; even if this were the way to go, you can catch `WebSocketException` specifically, and return the current state of the client.

VelvetThePanda (changes requested):

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
AraHaan: ConnectAsync requires the Uri endpoint which idk if it can be obtained from the old socket (it might be possible though).

AraHaan: I found a way to cache the endpoint by having ConnectAsync store it in an private field for this spot to use it.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -1047,6 +1047,16 @@ private async Task<Result> GatewayReceiverAsync(TimeSpan heartbeatInterval, Canc
...
+                    return Result.FromSuccess();
+                }
+
```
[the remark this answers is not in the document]
AraHaan: Done.

Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
AraHaan: Removed this catch.

Nihlus:
I'm not sure this is a good idea. It's best to propagate the state up to the caller and let them figure out if they need to give up or reconnect, potentially with a different endpoint. 

As it is, this PR will not be merged.

PR #213, opened 2022-06-25
Experimental rework of the interactivity system

Nihlus:
This PR reworks the interactivity system to use Remora.Commands instead of a half-baked homerolled system. It removes a lot of redundant code, fixes several issues with the previous implementation, and gives end users far more flexibility in custom implementations for interactions.

I'm opening this as a draft for now in order to get feedback on this and because I know others have worked on similar approaches - hopefully we can work together to make something we're all happy with :)


MazeXP:
Looks good to me so far.
I would propose to allow `[Ephemeral]` attribute as well like with Slash Commands:
https://github.com/Nihlus/Remora.Discord/blob/interaction-experimental/Remora.Discord.Commands/Responders/InteractionResponder.cs#L228-L240

Nihlus:
I believe new responses sent via interaction responses don't require this flag to be set anymore - can you try the system out and see if the behaviour matches what you'd expect?

AraHaan:
This rework is plan to have the same api structure that it has currently except the majority being implemented by Remora.Commands right?

VelvetThePanda:
> This rework is planned to have the same api structure that it has currently except the majority being implemented by Remora.Commands right?

No; this PR has a drastically different API than what could be considered "V1" of interactivity.

Samples/Interactivity/Commands/InteractiveCommands.cs
```diff
@@ -115,7 +118,7 @@ public async Task<IResult> SendDropdownAsync()
             {
                 new SelectMenuComponent
                 (
-                    "colour-dropdown",
+                    $"{Constants.InteractionTree}::{Constants.SelectMenuPrefix}::colour-dropdown",
```
AraHaan, line 121: I would actually like it if ``SelectMenuComponent`` prefixed this for us instead so we do not have to think about this ourselves on remembering to prefix it.

MazeXP:
> I believe new responses sent via interaction responses don't require this flag to be set anymore - can you try the system out and see if the behaviour matches what you'd expect?

I tested following scenario and it went quite well:
1. Slash Command `/interactive counter` to setup a base message with 2 buttons (Create and Create [Ephemeral]) for counter creation
2. Button creates a followup message either ephemeral or not ephemeral with a basic counter entity.
3. Each counter can be increased/decreased individually

While testing the scenario I came to the point that the class `InMemoryDataService` might be better part of `Remora.Discord.Interactivity` library, so it could also be reused by others.

I have tested as well the Pagination library and it works well for non-ephemeral messages.
In case of an ephemeral message it will not be able to edit the message unless either `EditOriginalInteractionResponseAsync` or `EditFollowupMessageAsync` instead of `EditMessageAsync´ is being used.
One just has to be cautious about caching in this case because the deletion of ephemeral is not able to be detected nor can be deleted by bot.

But all in all it feels good working with the new interactivity system.

AraHaan (approved):
Looks good to me, a comment however about SelectMenu interactions though.

Samples/Interactivity/Commands/InteractiveCommands.cs
MazeXP: I do not think it would be useful to have SelectMenuComponent do it as it is a base API object and should not be quite opinionated.
As well to have it not prefix so also other interactivity systems could be combined.

Remora.Discord.Interactivity/Attributes/SelectMenuAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SelectMenuAttribute"/> class.
+    /// </summary>
+    /// <param name="name">The menu's custom ID, excluding Remora's prefixed metadata.</param>
+    /// <param name="aliases">The menu's custom ID aliases, excluding Remora's prefixed metadata.</param>
+    public SelectMenuAttribute(string name, params string[] aliases)
+        : base
+        (
+            $"{Constants.SelectMenuPrefix}::{name}",
```
AraHaan, line 44: Is ``Constants.InteractionTree`` missing here?

Remora.Discord.Interactivity/Attributes/SelectMenuAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+            aliases.Select(a => $"{Constants.SelectMenuPrefix}::{a}").ToArray()
```
AraHaan, line 45: and here?

Remora.Discord.Interactivity/Attributes/SelectMenuAttribute.cs
MazeXP: It is not missing as `Constants.InteractionTree` is stripped from CustomID before being passed to `Remora.Commands`.

Samples/Interactivity/Commands/InteractiveCommands.cs
```diff
@@ -115,7 +118,7 @@ public async Task<IResult> SendDropdownAsync()
             {
```
AraHaan, line 118: If only there was an ``MessageComponentBuilder`` based class that acts like ``IServiceCollection`` where it's used to add "components" that would eventually build into the same thing, however it would probably need an nested ``ActionRowBuilder``, ``SelectMenuBuilder``, etc (and builders for the other interaction components), Also have ``MessageComponentBuilder`` have an ``Build()`` method responsible for calling ``Build()`` for the other builders which will new up all of those components in order.

I think the builder option for messagecomponents would be an great addition to the Interactivity apis too as an alternative to newing them up ourselves (the builders could be made to throw on improper setup of each type of component to tell the developer of their bot that they are not doing something properly).

Samples/Interactivity/Commands/InteractiveCommands.cs
```diff
@@ -115,7 +118,7 @@ public async Task<IResult> SendDropdownAsync()
...
                 new SelectMenuComponent
                 (
-                    "colour-dropdown",
+                    $"{Constants.InteractionTree}::{Constants.SelectMenuPrefix}::colour-dropdown",
                     new ISelectOption[]
                     {
                         new SelectOption("Red", "#FF0000"),
```
AraHaan, line 124: If an buider rout is added the SelectMenuBuilder could have an ``AddSelectOption`` method which will add the select options to an internal list that the ``Build()`` method would ``.ToArray()`` on when creating the ``SelectMenuComponent``.

Samples/Interactivity/Commands/InteractiveCommands.cs
Nihlus: Agreed; this rework is intended to allow coexistence with other systems.

Nihlus:
> > I believe new responses sent via interaction responses don't require this flag to be set anymore - can you try the system out and see if the behaviour matches what you'd expect?
> 
> I tested following scenario and it went quite well:
> 
> 1. Slash Command `/interactive counter` to setup a base message with 2 buttons (Create and Create [Ephemeral]) for counter creation
> 2. Button creates a followup message either ephemeral or not ephemeral with a basic counter entity.
> 3. Each counter can be increased/decreased individually
> 
> While testing the scenario I came to the point that the class `InMemoryDataService` might be better part of `Remora.Discord.Interactivity` library, so it could also be reused by others. As well as some utility class for generating the Custom IDs would be useful.
> 
> I have tested as well the Pagination library and it works well for non-ephemeral messages. In case of an ephemeral message it will not be able to edit the message unless either `EditOriginalInteractionResponseAsync` or `EditFollowupMessageAsync` instead of `EditMessageAsync´ is being used. One just has to be cautious about caching in this case because the deletion of ephemeral is not able to be detected nor can be deleted by bot.
> 
> But all in all it feels good working with the new interactivity system.

I've gone ahead and made pagination interaction-aware, which should alleviate most of this :) I'm not sure how Discord expects us to handle deletion of ephemeral messages, since you can't explicitly delete them via the REST API, nor do we get a notification if one has been deleted. For now, I'll avoid adding any explicit demonstrations of how to send ephemeral paginated messages - the way to do it is to suppress the automatic response and set the ephemeral flag on the contextual send.

I also added the suggested helper class for custom IDs. I'm still on the fence about `InMemoryDataService`, so I'll let others chime in as well.

MazeXP:
>Having a way to generate component sets from interaction groups would be nice. An extension method + an interface for more complex setups, maybe?
That sound good to be. Something like views?

>Is in-memory persistence something we want to ship with the library or let everyone figure out on their own? It's probably good to have a common, known-good implementation, but concurrency is tricky and not everyone has the same requirements.
It could be useful in my opinion as there might be some basic requirements for persisted data over interactions.
For more complex scenarios the user could still roll their own implementation.

> Should we add support (either here or in another PR) for sending interaction responses returned from command methods in order to save the user from having to manually send them if the initial response is suppressed?
I could think of some mechanism like ASP.Net Core with IResult, but I think that should be done in some other library itself and not the interactivity library.

PR #218, opened 2022-07-04
Externalize adding of Authorization header

MazeXP:
In the following the Authrization Header is now being added by a DelegatinHandler.
This allows to skip the addition of the Authorization header in special cases. (e.g. WebHooks)

An extension method has been added to the RestRequestBuilder class to skip the authorization. `b.SkipAuthorization()`

### DIFF
--- a/Backend/Remora.Discord.Rest/Extensions/RestRequestBuilderExtensions.cs
+++ b/Backend/Remora.Discord.Rest/Extensions/RestRequestBuilderExtensions.cs
@@ -62,14 +62,14 @@ public static class RestRequestBuilderExtensions
         bool isExemptFromGlobalLimits = false
     )
     {
-        var context = new Context
+        void ModifyContext(Context context)
         {
-            { "endpoint", builder.Endpoint },
-            { "cache", cache },
-            { "exempt-from-global-rate-limits", isExemptFromGlobalLimits }
-        };
+            context.Add("endpoint", builder.Endpoint);
+            context.Add("cache", cache);
+            context.Add("exempt-from-global-rate-limits", isExemptFromGlobalLimits);
+        }
 
-        builder.With(r => r.SetPolicyExecutionContext(context));
+        builder.With(r => r.ModifyPolicyExecutionContext(ModifyContext));
 
         return builder;
     }
@@ -83,7 +83,11 @@ public static class RestRequestBuilderExtensions
         this RestRequestBuilder builder
     )
     {
+#if NET5_0_OR_GREATER
+        builder.With(r => r.Options.Set(Constants.SkipAuthorizationOption, true));
+#else
         builder.With(r => r.Properties.Add(Constants.SkipAuthorizationPropertyName, true));
+#endif
 
         return builder;
     }
--- a/Backend/Remora.Discord.Rest/Extensions/RestRequestBuilderExtensions.cs
+++ b/Backend/Remora.Discord.Rest/Extensions/RestRequestBuilderExtensions.cs
@@ -73,4 +73,18 @@ public static class RestRequestBuilderExtensions
 
         return builder;
     }
+
+    /// <summary>
+    /// Adds an property to the request, used for skipping the addition of the Authorization header.
+    /// </summary>
+    /// <param name="builder">The request builder.</param>
+    /// <returns>The builder, with the property.</returns>
+    public static RestRequestBuilder SkipAuthorization(
+        this RestRequestBuilder builder
+    )
+    {
+        builder.With(r => r.Properties.Add(Constants.SkipAuthorizationPropertyName, true));
+
+        return builder;
+    }
 }
--- a/Backend/Remora.Discord.Rest/Extensions/ServiceCollectionExtensions.cs
+++ b/Backend/Remora.Discord.Rest/Extensions/ServiceCollectionExtensions.cs
@@ -40,6 +40,7 @@ using Remora.Discord.API.Objects;
 using Remora.Discord.Caching.Abstractions.Services;
 using Remora.Discord.Rest.API;
 using Remora.Discord.Rest.Caching;
+using Remora.Discord.Rest.Handlers;
 using Remora.Discord.Rest.Polly;
 using Remora.Rest;
 using Remora.Rest.Extensions;
@@ -74,7 +75,8 @@ public static class ServiceCollectionExtensions
         serviceCollection.ConfigureDiscordJsonConverters();
 
         serviceCollection
-            .AddSingleton<ITokenStore>(serviceProvider => new TokenStore(tokenFactory(serviceProvider)));
+            .AddSingleton<ITokenStore>(serviceProvider => new TokenStore(tokenFactory(serviceProvider)))
+            .AddSingleton<AuthorizationHandler>();
 
         serviceCollection.TryAddTransient<IDiscordRestAuditLogAPI>(s => new DiscordRestAuditLogAPI
         (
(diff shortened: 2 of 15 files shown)


Backend/Remora.Discord.Rest/Handlers/AuthorizationHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// <param name="tokenStore">The token store.</param>
+    public AuthorizationHandler(ITokenStore tokenStore)
+    {
+        _tokenStore = tokenStore;
+    }
+
+    /// <inheritdoc />
+    protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        if (!request.Properties.ContainsKey(Constants.SkipAuthorizationPropertyName))
```
AraHaan, line 50: Why is Properties used here over Options?

Backend/Remora.Discord.Rest/Extensions/RestRequestBuilderExtensions.cs
```diff
@@ -73,4 +73,18 @@ public static RestRequestBuilder WithRateLimitContext
...
+    /// <summary>
+    /// Adds an property to the request, used for skipping the addition of the Authorization header.
+    /// </summary>
+    /// <param name="builder">The request builder.</param>
+    /// <returns>The builder, with the property.</returns>
+    public static RestRequestBuilder SkipAuthorization(
+        this RestRequestBuilder builder
+    )
+    {
+        builder.With(r => r.Properties.Add(Constants.SkipAuthorizationPropertyName, true));
```
AraHaan, line 86: Why is Properties used here over Options?

MazeXP:
I could also add a TargetFramwork dependent preprocessor condition (`#if NET5_0_OR_GREATER`) to use for .NET 5 and later the Options variant.
As in .NET 5 the `HttpRequestMessage.Properties` has been deprecated and replaced by `HttpRequestMessage.Options`.

Nihlus:
This is nice, but there's another consideration that needs solving for this to work properly - rate limits are tied to the authorized entity as far as I know, and Remora has no concept of multiple concurrent rate limit sets at this time.

If you skip authentication and provide your own, you'll still be subject to existing rate limits, and could overwrite those with irrelevant data once your request completes. This would need a solution before  can merge this.

VelvetThePanda:
@Nihlus I could be reading this wrong, but to my understanding all this does is allow authorization to be *skipped*; no alternative auth is provided here. Afaik when it comes to the niche routes this helps fix (e.g. /webhooks), you generally wouldn't only skip auth on a route for some requests, so ratelimits should be handled like normal.

MazeXP:
@VelvetThePanda I understand his remarks.
The RateLimitPolicy does not differentiate between tokens (also no token) and does always count the call towards
the endpoint and global rate limit, but when no token is specified it should be actually tied to another rate limit context.

My solution would be to also have somehow the authorization header in the RateLimitPolicy, but I do not know of a
method to kinda forward a header to Polly context.

VelvetThePanda:
Ah, this is more situational, but already supported. `.IsExemptFromGlobalRatelimits(true)` if memory serves

Nihlus:
You could use the polly request context to pass that information into the rate limit handler; that said, such a change really needs exhaustive integration tests to be in place (which they are not, unfortunately)

MazeXP:
My idea for now would be to have global `RateLimitBucket`s per Authorization token and then
also have the endpoints `RateLimitBucket`s mapped per Authorization token + endpoint.

MazeXP:
According to my tests the endpoint rate limits does not differentiate individuals like the global rate limits.

AraHaan:
Is there anything else that needs to be done before this gets merged?

Nihlus:
Read back and you'll see.

MazeXP:
By integration test you mean the RestClient from start to end, not just the policy itself I suppose?

Backend/Remora.Discord.Rest/Handlers/AuthorizationHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Threading.Tasks;
+using Polly;
+using Remora.Discord.Rest.Extensions;
+
+namespace Remora.Discord.Rest.Handlers;
+
+/// <summary>
+/// Represents a delegating handler for adding the Authorization header.
+/// </summary>
+internal class AuthorizationHandler : DelegatingHandler
```
Nihlus, line 36: `BotTokenAuthorizationHandler` might be a more appropriate name for this type.

Tests/Remora.Discord.Rest.Tests/Polly/DiscordRateLimitPolicyTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        var cache = new MemoryCache(options);
+        return new MemoryCacheProvider(cache);
+    }
+
+    private static DiscordRateLimitPolicy CreatePolicy()
+    {
+        return DiscordRateLimitPolicy.Create();
+    }
+
+    private static Context CreateContext(
```
Nihlus, line 53: Formatting.

Nihlus:
Yeah, though I'm happy merging this with your handler-only tests. It's a big step in the right direction :)

Backend/Remora.Discord.Rest/Handlers/AuthorizationHandler.cs
MazeXP: I would actually prefer `TokenAuthorizationHandler`.
Also we might should add the TokenType to the ITokenStore, so that we might add also Bearer Token support if one wants to allow it to be used in OAuth context.

Nihlus: Sounds good to me!

Backend/Remora.Discord.Rest/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -56,13 +57,13 @@ public static class ServiceCollectionExtensions
...
     /// <param name="serviceCollection">The service collection.</param>
-    /// <param name="tokenFactory">A function that creates or retrieves the authorization token.</param>
+    /// <param name="tokenFactory">A function that creates or retrieves the authorization token and its token type.</param>
     /// <param name="buildClient">Extra client building operations.</param>
     /// <returns>The service collection, with the services added.</returns>
     public static IServiceCollection AddDiscordRest
     (
         this IServiceCollection serviceCollection,
-        Func<IServiceProvider, string> tokenFactory,
+        Func<IServiceProvider, (string Token, DiscordTokenType TokenType)> tokenFactory,
```
Nihlus, line 66: Why use a tuple here?

Tests/Remora.Discord.Commands.Tests/ContextTests.cs
```diff
@@ -52,7 +53,7 @@ public class ContextTests
     public ContextTests()
     {
         _services = new ServiceCollection()
-            .AddDiscordRest(_ => "dummy")
+            .AddDiscordRest(_ => ("dummy", DiscordTokenType.Bearer))
```
Nihlus, line 56: Slightly strange to use Bearer here, since Bot is the "standard" option. The use of Bearer implies that it has some effect on the test, which I don't believe is correct.

Tests/Remora.Discord.Commands.Tests/TestBases/CommandResponderTestBase.cs
```diff
@@ -46,7 +47,7 @@ public abstract class CommandResponderTestBase : IDisposable
     protected CommandResponderTestBase()
     {
         var serviceCollection = new ServiceCollection()
-            .AddDiscordRest(_ => "dummy")
+            .AddDiscordRest(_ => ("dummy", DiscordTokenType.Bearer))
```
Nihlus, line 50: Ditto.

Tests/Remora.Discord.Tests/Tests/Core/TokenStore/TokenStoreTests.cs
```diff
@@ -41,9 +41,23 @@ public class Token
         [Fact]
         public void ReturnsCorrectValue()
         {
-            var tokenStore = new TokenStore("Hello world!");
+            var tokenStore = new TokenStore("Hello world!", DiscordTokenType.Bearer);
```
Nihlus, line 44: Might want to have two separate tests, one for bearer and one for token.

Tests/Remora.Discord.Rest.Tests/TestBases/RestAPITestBase.cs
```diff
@@ -55,7 +55,7 @@ protected IServiceProvider CreateConfiguredAPIServices(Action<MockHttpMessageHan
         var serviceContainer = new ServiceCollection()
             .AddDiscordRest
             (
-                _ => "TEST_TOKEN",
+                _ => ("TEST_TOKEN", DiscordTokenType.Bearer),
```
Nihlus, line 58: Ditto.

Nihlus (changes requested):

PR #222, opened 2022-07-11
Allow Remora Plugins to add DiscordService using only service collection.

AraHaan:
Some people might prefer to use Remora.Plugins for providing all of the Discord Functionality in their programs. As such with this it gives them the ability to do so.

### DIFF
--- a/Remora.Discord.Hosting/Extensions/ServiceCollectionExtensions.cs
+++ b/Remora.Discord.Hosting/Extensions/ServiceCollectionExtensions.cs
@@ -1,0 +1,67 @@
+﻿//
+//  ServiceCollectionExtensions.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) 2017 Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


VelvetThePanda:
This PR is nonsensical; the point of `DiscordService` is to start the `DiscordGatewayClient` with a Generic Host.

AraHaan:
And they will still have the generic host, they might just want that all Discord code be in a Remora Plugin however (which includes it using ``AddDiscordService``). It is simply a pain to have them be forced to copy the code to the IHostBuilder version of it just to use it in their plugin.

VelvetThePanda:
Furthermore, DiscordService is a public class, so this PR is wholly unnecessary.

Remora.Discord.Hosting/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+[PublicAPI]
+public static class ServiceCollectionExtensions
+{
+    /// <summary>
+    /// Adds the required services for Remora Discord and a <see cref="IHostedService"/> implementation
+    /// using the service collection.
+    /// </summary>
+    /// <param name="serviceCollection">The service collection.</param>
+    /// <param name="tokenFactory">A function that retrieves the bot token.</param>
+    /// <returns>The service collection, with the services added.</returns>
```
AraHaan, line 47:
```suggestion
    /// <remarks>Only use this when the service collection is owned by an host builder.</remarks>
    /// <returns>The service collection, with the services added.</returns>
```

AraHaan:
An alternative is to change Remora Plugins where one can define plugins that need to configure host builders (which might be a breaking change to Remora Plugins).

VelvetThePanda:
*Or*, neither could happen, because no code needs to be changed. If you really want this behavior, you can just do it yourself; there's nothing special about Remora's methods.

PR #223, opened 2022-07-12
Pass target of User/Message context menu command to Command

MazeXP:
This PR allows to pass the TargetID of User/Message CommandTypes down to the Commands as named parameters.

Depending on the command type the TargetID is mapped as following:
```
ApplicationCommandType.User     =>  "user"
ApplicationCommandType.Message  =>  "message"
```

#### Usage
```cs
public class ContextMenuCommands : CommandGroup
{
    [Command("Handle Message")]
    [CommandType(ApplicationCommandType.Message)]
    public async Task<Result> HandleMessage(IMessage message)
    {
        // ...
        return Result.FromSuccess();
    }
    
    [Command("Handle User")]
    [CommandType(ApplicationCommandType.User)]
    public async Task<Result> HandleUser(IUser user)
    {
        // ...
        return Result.FromSuccess();
    }
}
```


Remora.Discord.Commands/Extensions/ApplicationCommandDataExtensions.cs
```diff
@@ -50,6 +50,32 @@ public static void UnpackInteraction
         out IReadOnlyDictionary<string, IReadOnlyList<string>> parameters
     )
     {
+        if (commandData.Type is ApplicationCommandType.User or ApplicationCommandType.Message)
+        {
+            var parameterName = commandData.Type switch
+            {
+                ApplicationCommandType.Message => "message",
+                ApplicationCommandType.User => "user",
+                _ => throw new InvalidOperationException()
```
AraHaan, line 59: This line should never be reached.

MazeXP: I would leave it there to avoid oversight when for example in the future there might be additional command types.
That might have some special handling.

MazeXP:
It might also be useful for some parser to make them InteractionContext aware as the Interaction object
contains the field "Resolved" with already resolved objects of type:
- IUser
- IPartialGuildMember
- IRole
- IPartialChannel
- IPartialMessage
- IAttachment

Remora.Discord.Commands/Extensions/ApplicationCommandDataExtensions.cs
Nihlus: Yes, having a throwing default is good practice and makes for safer code. Keep it as it is.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -479,13 +480,33 @@ ILocalizationProvider localizationProvider
...
+                        $"Command type {commandType} is not supported.",
+                        command
+                    );
+                }
+
+                if (parameters.Length != 1 || parameters[0].Name != expectedParameter)
+                {
+                    return new UnsupportedFeatureError
+                    (
+                        $"{commandType.Humanize()} context menu commands may only have a parameter named {expectedParameter}.",
```
Nihlus, line 506:
```suggestion
                        $"{commandType.Humanize()} context menu commands may only have a single parameter named {expectedParameter}.",
```

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -537,6 +558,12 @@ private static Result<IReadOnlyList<IApplicationCommandOption>> TryCreateCommand
         ILocalizationProvider localizationProvider
     )
     {
+        // Parameters options are only supported by slash commands
+        if (command.GetCommandType() is not ApplicationCommandType.ChatInput)
```
Nihlus, line 562: What's with this?

Remora.Discord.Commands/Parsers/MessageParser.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// <inheritdoc />
+    public override async ValueTask<Result<IMessage>> TryParseAsync
+    (
+        string value,
+        CancellationToken ct = default
+    )
+    {
+        if (!DiscordSnowflake.TryParse(value.Unmention(), out var messageID))
```
Nihlus, line 64: It's probably also useful to parse message links here, since those allow us to look up messages in channels other than the current context's channel.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
AraHaan: It's probably to try to guard against trying to parse any other parameters here as they are not supported for User/Message context commands?

Remora.Discord.Commands/Parsers/MessageParser.cs
MazeXP: By message link I suppose you mean these links?

In Guild: 
https://discord.com/channels/{GUILD_ID}/{CHANNEL_ID}/{MESSAGE_ID}
https://canary.discord.com/channels/{GUILD_ID}/{CHANNEL_ID}/{MESSAGE_ID}

In DM:
https://discord.com/channels/@me/{CHANNEL_ID}/{MESSAGE_ID}
https://canary.discord.com/channels/@me/{CHANNEL_ID}/{MESSAGE_ID}

I would actually implement the check for link using RegEx:
`https:\/\/(canary\.)?discord\.com\/channels\/(?<guild_id>@me|[0-9]*)\/(?<channel_id>[0-9]*)\/(?<message_id>[0-9]*)$`

Nihlus: Yep, that's correct - I'd use a compiled regex as well.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
MazeXP: I just moved the check in the method that is calling `TryCreateCommandParameterOptions`. Should be more cleaner.

As AraHaan just said the check is required so that additional parameters are not mapped to options as those are not supported by User/Message application commands.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -479,13 +480,33 @@ ILocalizationProvider localizationProvider
...
-                return new UnsupportedFeatureError
-                (
-                    "Context menu commands may not have parameters.",
-                    command
-                );
+                var expectedParameter = commandType switch
+                {
+                    ApplicationCommandType.Message => "message",
+                    ApplicationCommandType.User => "user",
+                    _ => null
```
Nihlus, line 490: Throw here instead.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -479,13 +480,33 @@ ILocalizationProvider localizationProvider
...
+                    return new UnsupportedFeatureError
+                    (
+                        $"Command type {commandType} is not supported.",
+                        command
+                    );
+                }
+
+                if (parameters.Length != 1 || parameters[0].Name != expectedParameter)
+                {
+                    return new UnsupportedFeatureError
```
Nihlus, line 504: This is not an unsupported feature; rather, it is a programmer error. Use a more appropriate error type.

Remora.Discord.Commands/Parsers/MessageParser.cs
AraHaan: regex source generator to the rescue here then.

uwx: ptb.discord.com also needs to be matched

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
AraHaan: Yeah throwing here would reduce the code right below this (since it will throw anyway since this sets it to null). It should help optimize because then it avoids an if which in this context would always be true when it returns null.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -479,20 +480,47 @@ ILocalizationProvider localizationProvider
...
             {
-                return new UnsupportedFeatureError
-                (
-                    "Context menu commands may not have parameters.",
-                    command
-                );
+                var expectedParameter = commandType switch
+                {
+                    ApplicationCommandType.Message => "message",
+                    ApplicationCommandType.User => "user",
```
AraHaan, line 489: In this context, what if the user clicks on a message to bring up the user profile to use the command, would it be possible to obtain that message then?

MazeXP: No, it would not be possible as in the user command context the message is not available.
It chooses the channel, the command response is sent to, depending on the open channel.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
MazeXP: I mean we could actually throw here as it would be a developer error.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
AraHaan: Except when you reply in the command it actually somehow gets the message from where you right clicked them and then reply to it.

MazeXP: What do you mean by reply?

AraHaan: When you use ``IDiscordRestInteractionAPI.CreateFollowupMessageAsync`` in the user context command.

MazeXP: It does not reference at all the message you used to bring up the user profile to use the command.
`IDiscordRestInteractionAPI.CreateFollowupMessageAsync` is just a reply to the interaction itself.

MazeXP: I mean technically what you are referring to is actually a message command.
You can get the user of the message from the Command context itself.

AraHaan: Just now changed the command from ``User`` to Message, waiting on when this can be merged now 👌.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -483,26 +483,10 @@ ILocalizationProvider localizationProvider
...
-                        command
-                    );
-                }
-
+                var expectedParameter = commandType.AsParameterName();
                 if (parameters.Length != 1 || parameters[0].Name != expectedParameter)
                 {
-                    return new UnsupportedFeatureError
-                    (
+                    return new InvalidParameterError(
```
Nihlus, line 489: Formatting.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -483,26 +483,10 @@ ILocalizationProvider localizationProvider
...
                         $"{commandType.Humanize()} context menu commands may only have a single parameter named {expectedParameter}.",
```
Nihlus, line 490: Also, should this be an error? It's a developer issue, which should generally raise an exception.

MazeXP: It is definitely a developer issue. Changed it to throw an `InvalidOperationException`.

Remora.Discord.Commands/Parsers/MessageParser.cs
MazeXP: Added handling of message links in `MessageParser`

Tests/Remora.Discord.Commands.Tests/Parsers/MessageParserTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                "TestUser",
+                1234,
+                default
+            )
+        );
+
+        var channelAPIMock = new Mock<IDiscordRestChannelAPI>();
+        channelAPIMock
+            .Setup(api =>
+                api.GetChannelMessageAsync(
```
Nihlus, line 69: Formatting.

Remora.Discord.Commands/Results/InvalidParameterError.cs
```diff
@@ -0,0 +1,33 @@
...
+using Remora.Commands.Trees.Nodes;
+using Remora.Results;
+
+namespace Remora.Discord.Commands.Results;
+
+/// <summary>
+/// Represents a failure to create a message/user command based on an invalid parameter.
+/// </summary>
+[PublicAPI]
+public record InvalidParameterError(string Message, IChildNode? Node = default) : ResultError(Message);
```
Nihlus, line 33: Is this type still used?

PR #227, opened 2022-07-23
WIP: Throw exceptions when a programmer has done something wrong

Nihlus:
This PR is part of a long-overdue comb-through of the library, wherein issues arising from programmer error (invalid API usage, not reading docs, etc) and not user input should raise exceptions instead of returning results.

I'm opening this as a draft PR in order to get community feedback on this more hardline approach, which is - at least theoretically - already in use in the library today. I've been relatively sloppy with this, and I need some extra eyes on this in order to get things reasonably right.


VelvetThePanda (approved):
I'll take a pass-through at some point; interactivity comes to mind regarding this, especially with some of the restrictions (e.g. command names being too long, misnamed parameters, etc.)

Hamsterland:
I appreciate all the effort with this PR, but I disagree with these changes; either because I am misinformed as to what their benefits are, or because I strongly believed Result usage should be unilateral. 

However, I am open-minded to this. What would you say is an example of "invalid API usage"? Regardless, if it is the developer who makes a usage mistake, then surely they'd rather be told that in a Result which they can handle gracefully, rather than an exception?

VelvetThePanda:
I think they're a welcome change and brings the library more in line with the [well documented philosophy](https://github.com/Nihlus/Remora.Discord/blob/master/.github/CONTRIBUTING.md#12-robustness)

Nihlus:
The core idea is to fail fast when you, as the developer, have done something that cannot be handled gracefully - the best comparison I have is C++'s `std::terminate()` function or Rust's `panic!()` macro. The class of problems that should result in exceptions are those of blatant API misuse or bugs, not those that could arise from normal and expected use.

VelvetThePanda:
Furthermore programmer error is rather exceptional, and should be treated as such, whereas with APIs and commands, failure is only a possibility when something goes wrong, not a guarantee. 

Most of these changes are exactly that, among with the areas I mentioned above.

Nihlus:
Rust, for all the things I fault with its syntax choices, explains the difference really well here: https://doc.rust-lang.org/std/macro.panic.html#when-to-use-panic-vs-result

Hamsterland:
Okay, I think I understand what you mean now. The rust `panic!()` macro put things into perspective. If I understand correctly, exceptions should only be thrown when the developer as grossly misused something in a way it was not intended to be used, and Results are returned from operational calls that have been used correctly. I agree with this philosophy, sorry I misunderstood.

Nihlus:
Exactly right :slightly_smiling_face:

AraHaan:
I disagree with these changes, if you want the developer to know they did something wrong that should be the job for an analyzer.

VelvetThePanda:
- Analyzers aren't available in every environment
- Analyzers require a lot of time investment for something as intricate as a Discord bot, and would require *constant* updating with the nature of a library like this
- While analyzers are useful tools, they're not a substitute for A) People that know the library making certain assertions and B) having runtime failures that can be picked up in CI/Unit tests, whereas analyzers require a warning/error being emitted. 
-

VelvetThePanda:
And given the already monumental workload Jax has to handle, I don't see an analyzer becoming an official solution in the foreseeable future, even if it weren't for the aforementioned caveats.

AraHaan:
Actually since I know how to make analyzers, I think I could get started on one.

VelvetThePanda:
*Furthermore* An analyzer would be another dependency added that's constantly running over code just for very niche circumstances, further degrading its necessity. I don't think it's a good idea.

Nihlus:
Having an analyzer would be a great compliment, but not a replacement.

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -804,31 +746,44 @@ private static Result ValidateNodeDescription(string description, IChildNode nod
...
+                    (
+                        $"A command description was too long (length {description.Length}, "
+                        + $"max {_maxCommandDescriptionLength}).",
+                        node
+                    );
+                }
+
+                if (description == string.Empty)
+                {
+                    return;
```
AraHaan, line 764: I think for non-context commands, having an empty description should throw.

MazeXP: You need to get more into detail here.
As this branch is actually enforcing an empty description for `USER` and `MESSAGE` commands.

Nihlus: Discord requires the description to be empty (i.e, not present) for `USER` and `MESSAGE` commands. Can you clarify what you mean?

MazeXP: AraHaan is actually referring to non-context commands (e.g. `CHAT_INPUT`).
According to Discord it must have a description with 1-100 characters, therefore empty string is actually invalid.

Nihlus: Right, which is checked in the conditional above this one.. This check is only for commands that are not chat input.

MazeXP: It checks if it is >100 characters but not if the description is empty.

Nihlus: That is true; however, Remora.Commands does not permit an empty description, and will substitude it with `No description set.` if one is encountered. This code can never see an empty description unless it's been explicitly overridden by Remora.Discord.Commands, which is the case for USER and MESSAGE commands (line 516).

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
```diff
@@ -237,21 +224,23 @@ public static Result<IReadOnlyList<IBulkApplicationCommandData>> CreateApplicati
         // Perform validations
         if (commands.Count > _maxRootCommandsOrGroups)
         {
-            return new UnsupportedFeatureError
+            throw new UnsupportedFeatureException
             (
-                $"Too many root-level commands or groups (had {commands.Count}, max {_maxRootCommandsOrGroups})."
+                $"Too many root-level commands or groups (had {commands.Count}, max {_maxRootCommandsOrGroups}).",
```
VelvetThePanda, line 229: Nitpick: Isn't it more typical to have `Max: x, Found/Detected: y`?

VelvetThePanda (approved):
LGTM

Remora.Discord.Commands/Extensions/CommandTreeExtensions.cs
AraHaan: I agree, that would be more helpful.

Issue #210, opened 2022-06-21
Implement the OAuth2 authorization flow


Hamsterland:
**Description**

Currently [`IDiscordRestOAuth2API`](https://github.com/Nihlus/Remora.Discord/blob/master/Backend/Remora.Discord.API.Abstractions/API/Rest/IDiscordRestOAuth2API.cs) only has two endpoints
- [Get Current Bot Application Information](https://discord.com/developers/docs/topics/oauth2#get-current-bot-application-information) 
- [Get Current Authorization Information](https://discord.com/developers/docs/topics/oauth2#get-current-authorization-information).

These are, to be fair, the only explicitly documented endpoints under the OAuth2 topic.

However, I think that the basic OAuth2 authorization flow endpoints should also have their own methods.
- [`/authorize`](https://discord.com/developers/docs/topics/oauth2#shared-resources-oauth2-urls)
- [`/token`](https://discord.com/developers/docs/topics/oauth2#shared-resources-oauth2-urls)
- [`/token/revoke`](https://discord.com/developers/docs/topics/oauth2#shared-resources-oauth2-urls) 

**Why This is Needed**
Remora.Discord should implement the full Discord API. Extracting the authorization flow into their own methods will expand Remora.Discord's use cases by simplifying the steps taken to authorize a user (e.g. in web apps). 

**Alternatives Considered**
While there are other libraries that do this for you, such as [AspNet.Security.OAuth.Providers](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers), these are very abstracted, feature-rich, and heavily opinionated towards usage in fully-fledged websites. As such, they do not integrate well with the Remora ecosystem.

I am proposing a simple wrapper around the authorization flow endpoints that does not involve Remora.Discord hosting any webservers to catch authorization codes - this is up to the consumer. 

**Additional Details**
Discord has its own OAuth2 URL generator that is quite simple to replicate. There could be a helper method to create these URLs too. 

...

If this is an issue you think is worth implementing, I would be happy to open a PR.

Hamsterland:
As per the reply on Discord, I'll go ahead and PR this

MazeXP:
Just a note that came up:
- The method `/authorize` endpoint should be returning the built url with specified parameters and not fetch it at all

AraHaan:
How would the bot get the authorization token though if it does not redirect to a least a website?

Hamsterland:
[...]

PR #228, opened 2022-07-29
Implement OAuth2 authorization flow (retrieve/revoke token)

MazeXP:
This PR implements only the OAuth2 specific endpoints that make sense to be called by an application:
- Retrieve token: `oauth/token`
- Revoke token: `oauth/token/revoke`

The `oauth/authorize` endpoint is not to be called by an application but been opened by an actual user.
Therefore we could provide some builder for those URLs.

### Related issues
#210 (Missing mentioned builder for authorize URLs)

### DIFF
--- a/Backend/Remora.Discord.API.Abstractions/API/Objects/OAuth2/IAccessTokenInformation.cs
+++ b/Backend/Remora.Discord.API.Abstractions/API/Objects/OAuth2/IAccessTokenInformation.cs
@@ -1,0 +1,59 @@
+//
+//  IAccessTokenInformation.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) 2017 Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+using System;
+using JetBrains.Annotations;
+using Remora.Rest.Core;
+
+namespace Remora.Discord.API.Abstractions.Objects;
+
+/// <summary>
+/// Represents access token response information.
+/// </summary>
+[PublicAPI]
+public interface IAccessTokenInformation
+{
+    /// <summary>
+    /// Gets the access token.
+    /// </summary>
+    string AccessToken { get; }
+
+    /// <summary>
+    /// Gets the type of the access token.
+    /// </summary>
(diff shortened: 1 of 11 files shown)


Backend/Remora.Discord.API.Abstractions/API/Objects/OAuth2/IAccessTokenInformation.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    /// <summary>
+    /// Gets the access token.
+    /// </summary>
+    string AccessToken { get; }
+
+    /// <summary>
+    /// Gets the type of the access token.
+    /// </summary>
+    string TokenType { get; }
```
AraHaan, line 43: Shouldn't this be ``GrantType``?

MazeXP: No, it will according to Discord always be "Bearer"

Nihlus:
While this looks perfectly fine, I'm somewhat concerned about the scope leakage of adding OAuth support like this. These structures and endpoints are not defined by Discord, but rather follows the application-independent [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749) definition of OAuth flows. 

There are multiple existing libraries that handle this in an integrated fashion, and I'm not sure this is something Remora should handle. I'd like some discussion from the community before deciding one way or another, though.

Hamsterland:
Not all OAuth implementations are the same. The API I am working with now for example does not support scopes or revoking tokens and has its own custom parameter for authentication. Discord’s is also slightly specialised. Therefore, I strongly believe this should remain in the library. There are no other OAuth packages in the Remora ecosystem. And I remember that when the lack of the OAuth API in Remora.Discord was brought up, there were more than a few of us who expected it to be here but were surprised that it wasn’t.

MazeXP:
I totally understand the point of Nihlus. As he is completely right that the OAuth definition is actually independent of Discord.
But I think all the existing OAuth client solutions have a hard dependency on OWIN and focus on the complete OAuth2 flow.
That means they do not support revoking the actual token. (e.g. on flogout the token should be revoked)
We could also just put it into its own discord independent project.

@Hamsterland
I suppose you mean `code_verifier`, `code_challenge` and `code_challenge_method`?
If yes, then those are also part of the OAuth definition under the name of PKCE. [[RFC](https://datatracker.ietf.org/doc/html/rfc7636)]
Discord does actually also support PKCE, but they have not mentioned it in their docs. It was only mentioned by night in an issue.

AraHaan:
I am personally indecisive if this is actually needed or not. For me personally I see no value in using it in my current discord bot however.

MazeXP:
The use case would mostly be desktop applications that allow users to authenticate using Discord.

Nihlus:
Based on this, I think it's reasonable we include it in Remora.Discord, then. However, the PR will need more work to properly handle OAuth error responses (probably by registering a new HttpClient just for these operations) and any additional considerations resulting from supporting what is ostensibly a different API surface than the Discord one.

MazeXP:
I would actually suggest then to have it in its own project not related to Discord:
    Remora.Reat.OAuth2

Therefore it could be a lightweight generalized implementation of the OAuth2 specs.

Hamsterland:
I think I agree with @MazeXP. The main concern here is that no other Auth library is part of the Remora ecosystem.

Nihlus:
I'll close this PR, then, and start putting together an OAuth library.

PR #230, opened 2022-08-26
Improve GatewayError printing

MazeXP:
I just extended the GatewayError handling in the DiscordGatewayClient to also
print the inner error to console.

This shows for example the cause why the gateway endpoint was not able to get.

### DIFF
--- a/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
+++ b/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
@@ -402,6 +402,11 @@ public class DiscordGatewayClient : IDisposable
                     message
                 );
 
+                if (iterationResult.Inner is { IsSuccess: false })
+                {
+                    _log.LogError("{Error}", iterationResult.Inner.Error);
+                }
+
                 shouldTerminate = true;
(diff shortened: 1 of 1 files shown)


Nihlus:
This is good, but I think a better option is to improve the pretty-printing of all errors in Remora.Results, then just use that.

MazeXP:
I see your point.
I suppose printing the error chain associated with a result would be to first to tackle.

My idea for pretty printing would be as follows:
```cs
Result? result = ...;
int indentLevel = 0
while (result is not null && !result.IsSuccess) {
    foreach (var property in result.Error.Properties) { // Properties = Public Properties of Error Type
        PrintIdented(indentLevel, $"{property.Name} = {property.Value}");
    }
    result = result.Inner;
    indentLevel++;
}
```

AraHaan:
I agree, it would be nice if Remora.Results could pretty print all errors for us to our logs.

Nihlus:
> 

Yep, this sounds good. Can you open an issue on Remora.Results and reference it here?

Nihlus:
Going to close this and refocus the idea onto Remora.Results.

PR #233, opened 2022-09-13
Add `NodeInjectionService` (#153, solution 2)

### LINKED ISSUE #153: Discussion: Future of command contexts and making contextual information more verbose
Myself personally, and I'd assume some others have stumbled upon the issue of not being able to detremine what really happened upon   command execution. 

@Nihlus pointed out three distinct solutions that all have various pros in cons and regards to how to solve this issue.

### Solution 1: Expanding `ICommandContext`

This one is a simple change, and would add an `IChildNode` or the likes to the pre-existing `ICommandContext` and it's derivitives.

This would more than likely be an `Optional<T>` property, seeing as contexts (namely `InteractionContext`) have had their usage scope expanded to more than just traditional commands (re: Interactivity)

### Solution 2: Node injection service

This change is also relatively simple, and piggybacks off a similar system that's currently in use by the command responders.

A new service would be added, by the names of `NodeInjectionService` or the likes. It will posses the current command node being invoked, much like `ContextInjectionService`.

## Solution 3: Reworking context architecture

This solution is largest of the three, and involves reworking how contexts are handled and constructed. 

Concrete types would posess a node property only if applicable, being omitted from the others.

`ICommandContext` would be minimally affected by this, as new base classes would be rewritten to derive it, especially in the case of interactions, where an `IInteractionContext` class would be spawned, and `InteractionContext` and `InteractionCommandContext` would derive from the former.

VelvetThePanda:
This PR implements support for solution 2 mentioned in #153.

I've yet to add tests (should be fairly simple). Solution 1 *may* have been easier considering interactivity being rewritten to piggyback off of Remora.Commands, but I already had the code written, and hadn't pushed. 

If solution one is preferred, 90% of the code can be reused so feel free to poke about that. cc @Nihlus

( Fixes #153 )

### DIFF
--- a/Remora.Discord.Commands/Extensions/CommandServiceExtensions.cs
+++ b/Remora.Discord.Commands/Extensions/CommandServiceExtensions.cs
@@ -1,0 +1,88 @@
+//
+//  CommandServiceExtensions.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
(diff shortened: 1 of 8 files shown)


VelvetThePanda:
Fixes #153

Nihlus:
I'm kind of still in the solution 3 camp, myself. That said, this is not orthogonal to that and could be merged as a stopgap measure.

VelvetThePanda:
The third option is probably the best? Certainly a fair bit more work to implement, so I second this being a stopgap.

Nihlus:
You've got some xUnit warnings to take care of.

Nihlus:
Closing as moot.

PR #239, opened 2022-10-05
Guard GetRequiredService call in try-catch in DispatchEventAsync

uwx:
Normally, The call to IServiceProvider.GetRequiredService in ResponderDispatchService.DispatchEventAsync isn't guarded by a try-catch; however, GetService can throw an exception if the service activation fails for some reason. This causes an exception to be absorbed instead of rethrown as an Error (which exceptions thrown by RespondAsync _will_).

This PR moves the call to GetRequiredService into the `try`.

This PR does not fix the underlying problem of unguarded exceptions within DispatchEventAsync being swallowed; it is unclear to me what should happen in that case, however I believe putting the GetRequiredService call inside the `try` is still the correct fix for the immediate problem.

### DIFF
--- a/Backend/Remora.Discord.Gateway/Services/ResponderDispatchService.cs
+++ b/Backend/Remora.Discord.Gateway/Services/ResponderDispatchService.cs
@@ -326,11 +326,11 @@ public class ResponderDispatchService : IAsyncDisposable
                     async rt =>
                     {
                         await using var serviceScope = _services.CreateAsyncScope();
-                        var responder = (IResponder<TGatewayEvent>)serviceScope.ServiceProvider
-                            .GetRequiredService(rt);
-
                         try
                         {
+                            var responder = (IResponder<TGatewayEvent>)serviceScope.ServiceProvider
+                                .GetRequiredService(rt);
+
                             return await responder.RespondAsync(gatewayEvent.Data, ct);
                         }
                         catch (Exception e)


MazeXP:
It is totally fine, that GetRequiredService would throw an exception.
To my extent it is a programmers error, as some service was not being registered properly with DI.
ResultErrors should only be used when it is not caused by misuse of the library by programmers.

Do you have any example that shows an exception throwing when it is clearly not a programmers misuse?

Nihlus:
Yeah, this is intentional behaviour. A better fix would be to ensure that the service is added as part of the outer service addition call.

uwx:
@MazeXP @Nihlus Okay, so I should better explain my reasoning for this change.

When GetRequiredService is _not_ guarded, and the call throws, several things happen
- GetRequiredService throws an InvalidOperationException (or other exception)
- The `await using` call above causes an async `finally` to occur, which transitions from synchronous to asynchronous execution
- The Task is transitioned into an errored state
- `await Task.WhenAll` bubbles up the exception; **all subsequent responders for that gateway event will not execute, and DispatchEventAsync resolves to faulted Task**
- responderTask in UnwrapAndDispatchEvent faults
- The task, by now, was already sent to `_respondersToFinalize`
- FinalizerTaskAsync does `FinalizeResponderDispatch(await responderResults);`; **Here it's awaiting a faulted Task, which throws an exception, which it cannot recover from**
- At this point, the dispatcher is fine (it does care about the Task's result), but the finalizer has crashed; as finalizer faults are not handled in any way, **the finalizer is dead until program restart and the user is not aware of this; the library is irrevocably broken. There is no exception logged, and the program does not fully crash**

So, let me list out the issues here that need addressing, of which my PR addresses the first of them:
- An exception in GetRequiredService in the dispatcher is not safely handled, and will cause DispatchEventAsync to fault, which means all subsequent dispatchers for that event will fail to execute
- An exception in DispatchEventAsync (more specifically the dispatchDelegate call in UnwrapAndDispatchEvent) will not trigger any of the failsafes:
  - The try-catch for `responder.RespondAsync` will successfully catch exceptions created by the responder itself, but not by the service activation, because the GetRequiredService call is unguarded
  - `Failed to dispatch payload: {Reason}` is not logged, because this exception is asynchronous
  - `Error in gateway event responder: {Exception}` is not triggered, because that code is never reached; the code fails when `Task<IReadOnlyList<Result>>? responderResults` is a faulted task, as it's directly awaited with no error handling; FinalizerTaskAsync does not expect that task to be faulted, it only logs a possible failed Result inside it
  - `AppDomain.CurrentDomain.UnhandledException` is never invoked; it's unclear to me as to why, but this shouldn't be relied upon anyway

For further reference, here are the results in console of an exception being thrown from within GetRequiredService with and without this PR:

Without:
rider64_2022-10-08_13-30-53

With:
rider64_2022-10-08_13-32-33


---

> A better fix would be to ensure that the service is added as part of the outer service addition call.

Could you explain what you meant by this?

Nihlus:
Aha, I see. In that case, let me reopen this.

What I meant by that comment is that an error in GetRequiredService should not be an error that is gracefully handled; rather, it is a programmer error of some kind which should bubble up during development. However, given the detailed explanation of the issue and its subsequent undesired effects, this is a fair fix.

PR #241, opened 2022-10-06
Improve memory pressure in ImagePacker by not allocating extra byte[] buffers if they can be retrieved from the Stream directly

uwx:
ImagePacker.PackImageAsync creates a new `byte[]` out of the Stream even if a MemoryStream is passed; if the input stream is not a MemoryStream, it allocates *two* byte arrays with the size of the image being uploaded.

This PR fixes this by using TryGetBuffer if the underlying Stream is a MemoryBuffer and allows it, falling back to ToArray otherwise. When the Stream has to be copied to a buffer, we also retrieve the MemoryStream's underlying buffer instead of allocating a new one.

### DIFF
--- a/Backend/Remora.Discord.Rest/Utility/ImagePacker.cs
+++ b/Backend/Remora.Discord.Rest/Utility/ImagePacker.cs
@@ -85,35 +85,47 @@ public static class ImagePacker
         CancellationToken ct = default
     )
     {
-        byte[] bytes;
+        ReadOnlyMemory<byte> bytes; // Must be a Memory, as this method is async
         if (stream is MemoryStream ms)
         {
-            bytes = ms.ToArray();
+            bytes = GetBufferOrToArray(ms);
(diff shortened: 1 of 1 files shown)


uwx:
Depends on https://github.com/Remora/Remora.Rest/pull/15

uwx:
Will undraft this PR when I test it.

uwx:
Tested by uploading a bunch of custom emoji and it's working fine.

Nihlus:
You've got some build errors you need to take care of.

uwx:
Build error is due to the dependency on https://github.com/Remora/Remora.Rest/pull/15, which is not yet on NuGet, so there's not much I can do there.

Nihlus:
Remora.Rest has now been updated. Could you rebase this PR onto the latest main branch?

uwx:
Did that

PR #244, opened 2022-10-20
Make RedisCacheProvider pass values as byte[] directly instead of serializing to string

uwx:
The current implementation of `RedisCacheProvider.CacheAsync` serializes into a string, then passes that to the RedisCache; passing a byte array should be more friendly to the System.Text.Json serializer and reduce string processing overhead.

This PR uses `JsonSerializer.SerializeToUtf8Bytes` and the `byte[]` overload of `IDistributedCache.SetAsync` for this.

### DIFF
--- a/Backend/Remora.Discord.Caching.Redis/Services/RedisCacheProvider.cs
+++ b/Backend/Remora.Discord.Caching.Redis/Services/RedisCacheProvider.cs
@@ -75,7 +75,7 @@ public class RedisCacheProvider : ICacheProvider
             return;
         }
 
-        var serialized = JsonSerializer.Serialize(instance, _jsonOptions);
+        var serialized = JsonSerializer.SerializeToUtf8Bytes(instance, _jsonOptions);
 
         var options = new DistributedCacheEntryOptions
         {
@@ -83,7 +83,7 @@ public class RedisCacheProvider : ICacheProvider
             SlidingExpiration = slidingExpiration
         };
 
-        await _cache.SetStringAsync(key, serialized, options, ct);
+        await _cache.SetAsync(key, serialized, options, ct);
     }
 
     /// <inheritdoc cref="ICacheProvider.RetrieveAsync{TInstance}"/>


VelvetThePanda:
This is a good change in theory! However, I'm struggling to benchmark performance improvements. 

Sometimes it's a 20% gain:

```
|         Method |       Mean |    Error |    StdDev |   Median | Allocated |
|--------------- |-----------:|---------:|----------:|---------:|----------:|
|         Remora | 1,064.6 us | 68.20 us | 321.85 us | 941.7 us |         - |
| RemoraImproved |   861.1 us | 20.16 us |  94.34 us | 852.3 us |         - |
```

Sometimes it's __<__ 1%:
```
|         Method |     Mean |    Error |    StdDev |   Median | Allocated |
|--------------- |---------:|---------:|----------:|---------:|----------:|
|         Remora | 827.5 us | 10.52 us |  99.02 us | 813.7 us |         - |
| RemoraImproved | 837.9 us | 11.43 us | 108.06 us | 816.6 us |         - |

```

VelvetThePanda:
I've also just now realized that I'm using the synchronous methods, but I suspect that the overhead of async/await is going to crush any tangible differences

VelvetThePanda:
Aha. I upped the op count to 100, and now we can see some more viable information:

```
|         Method |     Mean |    Error |   StdDev | Allocated |
|--------------- |---------:|---------:|---------:|----------:|
|         Remora | 748.1 us |  9.99 us | 10.26 us |   1,534 B |
| RemoraImproved | 758.0 us | 14.68 us | 21.06 us |     969 B |
```

VelvetThePanda:
In even larger tests (~1500 `op`), this PR actually ends up allocating *more*, but again, perhaps I'm benchmarking wrong. 

```
|         Method |     Mean |    Error |   StdDev | Allocated |
|--------------- |---------:|---------:|---------:|----------:|
|         Remora | 753.9 us |  7.72 us |  6.44 us |      1 KB |
| RemoraImproved | 762.2 us | 11.77 us | 10.43 us |      2 KB |
```

uwx:
Evidently there are too many moving parts in the whole Redis cache process to easily benchmark it. The gains are probably too small to be noticeable against the overhead of everything else going on. As Redis integration does a whole lot of I/O, that is likely the bottleneck.

If you really want to have some performance figures, perhaps the best you can do is benchmark `Serialize` vs `SerializeToUtf8Bytes`.

VelvetThePanda:
```
|         Method |     Mean |    Error |   StdDev | Allocated |
|--------------- |---------:|---------:|---------:|----------:|
|         Remora | 807.7 ns | 16.23 ns | 33.87 ns |      73 B |
| RemoraImproved | 691.8 ns | 13.41 ns | 27.10 ns |      49 B |
```

uwx:
There you go, marginally faster :P

PR #245, opened 2022-10-21
Refactor API of ICacheProvider to not unnecessarily duplicate MemoryCacheEntryOptions/DistributedCacheEntryOptions instances

uwx:
This PR changes the ICacheProvider API to use a new CacheEntryOptions type as the least common denominator for MemoryCacheEntryOptions and DistributedCacheEntryOptions.

## Motivation
* Both current cache providers use either MemoryCacheEntryOptions or DistributedCacheEntryOptions, but both of them create the instances on every call to ICacheProvider.CacheAsync.
* Worse yet, CacheInstanceAsync calls CacheSettings.GetEntryOptions, which creates a MemoryCacheEntryOptions instance of its own!
* Evidently, the APIs aren't being used properly here.
* None of these cache-entry-options are cached either, which seems quite funny for a cache service.
  * The `Microsoft.Extensions.Caching` APIs don't mutate the entry options (they copy them into their internal entity structure), which allows them to be shared across all cache entries that are configured the same.
* There are likely to be performance improvements on allocation here, but they're small compared to things like I/O overhead to deserialize the to-be-cached objects; Instead you should see this PR as a fix for API correctness rather than to improve performance.

## Noteworthy changes
* A CacheEntryOptions instance is added, because as mentioned above there is no common ancestor for MemoryCacheEntryOptions and DistributedCacheEntryOptions.
  * This instance lazily initializes either MemoryCacheEntryOptions or DistributedCacheEntryOptions instances, or both, as desired, via implicit operators that cache the result. This is the simplest and least problem-laden solution.
* The CacheSettings entity is no longer used directly, instead we have ImmutableCacheSettings.
  * This is because the computed CacheEntryOptions might not remain correct if a default configuration is set after an entity-specific configuration; this would make calls to different configuration methods order-interdependent, which is Very Bad.
  * The ImmutableCacheSettings is a singleton that takes the CacheSettings and caches all necessary CacheEntryOptions instances ahead of time. This is the simplest and least problem-laden solution.
  * This means a new service is added.
* The ICacheProvider API is adjusted to accept the new CacheEntryOptions object. This change would be necessary anyway, for the bugfix mentioned in the section below.
* `Remora.Discord.Caching.Abstractions` gets a dependency on `Microsoft.Extensions.Caching.Abstractions`. No big deal, something was probably going to get this dependency anyway.

## Bugs in the existing implementation
* CacheSettings only allows for defining AbsoluteExpirationRelativeToNow (TimeSpan) and not AbsoluteExpiration (DateTimeOffset), but ICacheProvider only uses AbsoluteExpiration (DateTimeOffset). This means setting an absolute expiration doesn't work at all, because the TimeSpan and DateTimeOffset properties are not interchangeable. This PR fixes it by passing the cache entry options as-is in the ICacheProvider, and relaying them to the internal IMemoryCache/IDistributedCache.

## RFC
Unless there are unit tests for the caching system I haven't found, I'll need a willing volunteer to test the Redis changes, at least.


uwx:
Tests added via #246 pass for this PR, only Redis testing is required now.

Backend/Remora.Discord.Caching.Abstractions/CacheEntryOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// Gets the absolute expiration time, relative to now.
+    /// </summary>
+    public TimeSpan? AbsoluteExpirationRelativeToNow { get; init; }
+
+    /// <summary>
+    /// Gets the sliding expiration time.
+    /// </summary>
+    public TimeSpan? SlidingExpiration { get; init; }
+
+    private MemoryCacheEntryOptions? _memoryCacheOptions;
```
Nihlus, line 52: Private fields should be declared before public properties.

Backend/Remora.Discord.Caching.Abstractions/CacheEntryOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// <summary>
+    /// Gets the cached <see cref="MemoryCacheEntryOptions"/> for this instance, or creates a new one and caches it.
+    /// </summary>
+    /// <param name="self">The current <see cref="CacheEntryOptions"/> instance.</param>
+    /// <returns>
+    /// An instance of <see cref="MemoryCacheEntryOptions"/> with settings corresponding to the current instance.
+    /// </returns>
+    public static implicit operator MemoryCacheEntryOptions(CacheEntryOptions self)
+    {
+        return self._memoryCacheOptions ??= MakeOptions(self);
```
Nihlus, line 64: Could we use a Lazy<T> for this instead?

Backend/Remora.Discord.Caching/Services/CacheSettings.cs
```diff
@@ -23,56 +23,67 @@
...
-using Microsoft.Extensions.Caching.Distributed;
-using Microsoft.Extensions.Caching.Memory;
 
 namespace Remora.Discord.Caching.Services;
 
 /// <summary>
 /// Holds various settings for individual cache objects.
 /// </summary>
+/// <remarks>
+/// Cache services should use <see cref="ImmutableCacheSettings"/> instead.
```
Nihlus, line 33: Is this type deprecated due to this? If so, we should probably just delete it outright since the PR contains API breaking changes anyway.

Backend/Remora.Discord.Rest/Polly/DiscordRateLimitPolicy.cs
```diff
@@ -162,6 +163,9 @@ bool continueOnCapturedContext
             return response;
         }
 
+        // TODO does moving this cacheOptions up here cause regressions?
```
Nihlus, line 166: Does it? :P

Backend/Remora.Discord.Caching/Services/CacheSettings.cs
uwx: No, since the ImmutableCacheSettings derives from CacheSettings. The user configures CacheSettings via DI, but cache services themselves consuem the Immutable one.

Backend/Remora.Discord.Caching.Abstractions/CacheEntryOptions.cs
uwx: Yes, but a Lazy has some poignant characteristics that make it likely unfit for this use:

- It's thread-safe by default, which we don't need
- It guarantees only single initialization, which we don't need
- It uses an annoyingly large amount of memory per instance due to storing delegates and such (40 bytes if I recall? Which when we have to have two of them, is slightly painful)

If you still wish to use a Lazy for this, I can do that, but this is in my opinion one of the situations where it's not applicable.

Nihlus:
I'm happy merging this after you fix the remaining change requests and merge conflicts :)

Backend/Remora.Discord.Rest/Polly/DiscordRateLimitPolicy.cs
uwx: On second pass, no reason why it should.

uwx:
Tests from #246 pass for this PR.

PR #246, opened 2022-10-22
Add tests for Remora.Discord.Caching

uwx:
Adds unit tests for basic functionality of the cache system. Note that these tests **will not pass** as the cache configuration functionality is broken, which is fixed by #245.

PR #247, opened 2022-10-22
Create a value-based cache key system

uwx:
This PR changes the string-based cache key system in Remora.Discord.Caching to one where each distinct type of cache key receives its own type, and uses record equality to compare that cache keys must be from the same type to be equal.

## Motivation
* The IMemoryCache allows arbitrary objects to be passed as keys. The current system uses strings only, which is relatively flaky: If a key parameter contains `:`, it breaks the semantics of the key, as `:` is used as a separator control character.
* It's also potentially (in theory) wasteful to allocate and format entire strings every time an entry is to be sent to or pulled from the cache.
* A char-by-char comparison with strings is possibly less efficient than a straight up value comparison, especially when most of the values are Snowflakes, which get serialized as large numbers but are otherwise just 64-bit values.
* You allowed me to get annoying about saving bytes in Snowflakes, so now I'm fully unhinged.

## Changes
* A new CacheKey base record is added to be the supertype of any kind of cache key.
* Being a record, it's compared for equality by type and value, meaning two identical records implementing CacheKey will **not** be ambiguous to the cache backend.
* A utility method, CacheKey.StringKey, is added for those who desperately need keys made out of plain strings.
* Another utility method, CacheKey.LocalizedStringKey, is added in order to provide the simplicity of string-based keys while also differentiating between two identical keys that come from different domains. One example of this is used by the Polly backend.
* CacheKey has a ToCanonicalString method that converts the key into a string representation for caches that do not support arbitrarily-typed keys. Note that this opts out of any safety guarantees that the record system adds.
* ICacheProvider's API is modified to take CacheKey keys instead of strings.
* Keys are now created by `new KeyHelpers.XXXXXCacheKey()`
* The Redis cache backend simply does `x.ToCanonicalString` on the CacheKey instances as IDistributedCache expects a string-based key.

## Notes
I'm creating this PR as a draft as it's likely to elicit discussion and debate. Feel free to leave your thoughts on this idea.


Nihlus:
The caching system used to use tuples with strong typing for the cache keys, but was switched over to strings in order to have a uniform key system between Redis and memory caching. I'd like to get the Redis users input on this before I make a determination one way or the other.

Nihlus:
@VelvetThePanda - any input, since you've been using Redis?

VelvetThePanda:
Seems fine enough to me?

Nihlus:
In that case, I'm on board with the basic concept of this change. Rebase/resolve the merge conflicts and lift the WIP marking and I'll get to proper reviewing :)

uwx:
Tests from #246 pass for this PR when the fixes from #245 are applied.

PR #256, opened 2022-11-20
Use global.json for SDK management

MazeXP:
Uses global.json mechanism to manage Remora.Sdk and Remora.Tests.Sdk version.

### DIFF
--- a/Backend/Remora.Discord.API.Abstractions/Remora.Discord.API.Abstractions.csproj
+++ b/Backend/Remora.Discord.API.Abstractions/Remora.Discord.API.Abstractions.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Remora.Sdk/2.0.1">
+<Project Sdk="Remora.Sdk">
 
     <PropertyGroup>
         <VersionPrefix>72.0.0</VersionPrefix>
--- a/Backend/Remora.Discord.API/Remora.Discord.API.csproj
+++ b/Backend/Remora.Discord.API/Remora.Discord.API.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Remora.Sdk/2.0.1">
+<Project Sdk="Remora.Sdk">
 
     <PropertyGroup>
         <VersionPrefix>68.0.0</VersionPrefix>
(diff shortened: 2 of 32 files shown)


Nihlus:
As it is, this is going to break any build whose working directory is not in the root of the repository. I'm not keen on doing that.

uwx:
> As it is, this is going to break any build whose working directory is not in the root of the repository. I'm not keen on doing that.

Does this only apply to people who directly reference Remora csprojs from their solutions? I do that, and I wouldn't really mind having to copy and paste the global.json into my solution rootdir. If necessary you can relicense that particular file under a more open license.

Nihlus:
That's one of the cases it applies to, but it'd also break a simple `dotnet build` in any of the individual projects.

MazeXP:
I was not able to reproduce failing `dotnet build` from any individual project and also `dotnet build <absolute_path_to_csproj>` from `/home/` did not fail either.
The only case that fails is as @uwx said is when directly referencing the project outside of the solution directory.

Nihlus:
Interesting - that doesn't map with what I read about global.json and how the projects have been acting before when a version wasn't explicitly specified in the csproj files themselves.

Could you try with a cleared cache and a fresh clone, just to make sure?

MazeXP:
Just tested with following .NET version: `7.0.100`

```
Clear all NuGet caches:
> dotnet nuget locals all -c

Clone repository:
> https://github.com/MazeXP/Remora.Discord.git
> git checkout use-global-json

Switch into Remora.Discord.Api directory and build:
> cd Backend/Remora.Discord.Api
> dotnet build
=> Restored NuGet dependencies
=> Built all 3 build targets with dependencies
```

and 

```
Clear all NuGet caches:
> dotnet nuget locals all -c

Clone repository:
> https://github.com/MazeXP/Remora.Discord.git
> git checkout use-global-json

Switch into /home/ and build with absolute path:
> cd /home/
> dotnet build "/tmp/Remora.Discord/Backend/Remora.Discord.Api/Remora.Discord.Api.csproj"
=> Restored NuGet dependencies
=> Built all 3 build targets with dependencies
```

No errors occured in this case.

Nihlus:
After some clarification on the Discord server, I'm happy with this change.

PR #264, opened 2022-12-20
Readd AddDiscordCaching call to Caching sample

MazeXP:
The AddDiscordCaching call was erroneously when it was first replaced by AddDiscordRedisCaching and the removed when removing the dependency of Redis.


Nihlus:
PR does not build.

MazeXP:
Was missing the `using`. Should be fine now.

PR #271, opened 2022-12-28
Allow passing state with CustomIDs

MazeXP:


Nihlus:
I'm not entirely sure about this approach. If you want to pass along state, you should probably do it via the in-memory data service + generated custom IDs. I recall the folks at Discord not being keen on people encoding state in custom IDs.

That said, I'm open to input and would like to hear from the rest of the community on this.

VelvetThePanda:
Discord has been pushing developers toward stateless bots ever since slash commands came out. The 100 character custom ID limit (while somewhat limited for some) is at least passively encouraged to be used to encode some semblance of state for the purpose of making stateless (HTTP) bots easier to develop.

MazeXP:
Even in the docs they mention to use custom IDs also for state: https://discord.com/developers/docs/interactions/message-components#custom-id

Nihlus:
Alright, in that case, I'm on board.

MazeXP:
I'm not quite sure on the prefix for the state to be honest, if it is unique enough or another approach would be better.
Also it would be also good to think about reducing the prefix used by Remora.Interactivity itself, to provide more space for state and so on.

Nihlus:
I'm not aware of any restrictions on the contents of `custom_id`, so we could theoretically start using some more exotic character values as our delimiters - ASCII GS/RS/US come to mind. It might even be worthwhile to explore some kind of compressed or encoded format for the whole thing and allow end users to provide serialized state data.

VelvetThePanda:
Fun fact: People tried encoding their custom IDs in Base1024 to cram as much data as possible into 100 characters

VelvetThePanda:
That being said, I think `U+241D` (Group Separator) would be quite apt.

MazeXP:
I mean for the first iteration we could just stay with basic string and in another iteration we could provide some.utility method to encode and compress some state object into a string.

Remora.Discord.Interactivity/Constants.cs
```diff
@@ -34,4 +34,9 @@ public static class Constants
     /// Gets the name of the interaction command tree.
     /// </summary>
     public const string InteractionTree = "remora::interactivity";
+
+    /// <summary>
+    /// Gets the prefix used for passing the state.
+    /// </summary>
+    public const string StatePrefix = "$!";
```
Nihlus, line 41: Let's swap this over to `0x1D` (ASCII Group Separator), since it's highly unlikely to appear in plaintext state information.

Nihlus: Actually, nevermind. It's best done later in a rework.

PR #272, opened 2022-12-31
Add GatewaySession and client APIs

VelvetThePanda:
This PR adds the ability to start the gateawy client with a pre-determined `_sessionID` and `_lastSequenceNumber` to enable using resumes instead of identifies. 

There's a default limit of 1000 IDENTIFY/day, and resumes replay events starting with the given `s` (sequence). 

[Identify Docs](https://discord.dev/topics/gateway#identifying)
[Resume Docs](https://discord.dev/topics/gateway#resuming)


Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -197,6 +197,50 @@ ResponderDispatchService responderDispatch
...
+    /// If the client reconnects, or is still running, the session ID and sequence number may be invalid.
+    /// </remarks>
+    /// <returns>A result containing the current gateway session information.</returns>
+    public Result<GatewaySessionInformation> GetGatewaySessionInformation()
+    {
+        if (string.IsNullOrEmpty(_sessionID) || _lastSequenceNumber is 0)
+        {
+            return Result<GatewaySessionInformation>.FromError
+            (
+                new InvalidOperationException("No session information is available.")
```
Nihlus, line 233: A `NotFoundError` might be more appropriate. Also, you should be able to just return the result directly.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
```diff
@@ -300,13 +344,11 @@ public async Task<Result> RunAsync(CancellationToken stopRequested)
         }
         finally
         {
-            _sessionID = null;
```
Nihlus, line 303: Should this not null out the session ID in the session information or replace the session information entirely?

Nihlus (changes requested):
Looks reasonable. Don't forgot about tests.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
VelvetThePanda: Elaborate?

Nihlus: Well, the whole point of this section is to reset the internal state to a clean slate when session resumption is explicitly not possible or desired. As the code is, the old session information will always hang around.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
VelvetThePanda: The thought process was that the gateway client always preserves its last known state with the gateway so that it can be extracted by the user in a safe manner (e.g. to be resumed after a startup)

VelvetThePanda: So, if this error is returned, the only possible state is that the client was never started, and thusly this method is being used incorrectly. So perhaps an exception, because of developer error.

Nihlus: I assume you wanted to reply to the other comment :P but, as I said, the points where the session ID is nulled out  are points where we know for a fact that the session is *not* resumable, and as such there's no point in keeping the information around. It's more likely to cause issues than anything.

Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs
VelvetThePanda: Right. In theory the only time that this should happen is when Discord sends `OP9` (Reconnect) with `d = false` (which indicates that the session cannot be resumed).

VelvetThePanda:
Gonna drop a comment as to not spam the review comments. 

As for causing issues, the client goes based off of `_isSessionResumable`, which should be set to false in the case of `OP9` with `d = false`, so there shouldn't be an issue with that as far as I can tell.

Nihlus:
Right, but my point is this - why preserve and expose a session ID to the user that is known to be invalid? It's just setting them up to hit an error the moment they try to use it, accidentally or otherwise.

VelvetThePanda:
The session is not immediately invalidated, actually. As per the resume docs I linked earlier:
> After the connection is closed, your app should open a new connection
This implies that there is a grace period between disconnecting and the session actually being terminated serverside. This is also how large bots achieve zero-downtime restarts.

Unfortunately, Discord isn't explicit in this timeframe, but it's generally understood to be "long enough" for a new client to be started (which is the typical use-case).

Nihlus:
Yes, but that's for *valid* session IDs, which are used in the current resume code, and would be fine if you pulled them out with the method you've added. I'm talking about the case where the session ID is explicitly invalidated by Discord and we set it to null to discard the now-bad data.

I realize I probably attached the original comment to the wrong line (the one that nulls out the session ID after a clean shutdown by user request, which may have contributed to some confusion. Only one of the two `= null` lines are relevant to this discussion - the one reached after an `OP9/false`.

VelvetThePanda:
So, I've taken a look through the gateway code to see where this could be an issue, and I think I've spotted several more issues (namely with the impossible handling of reconnects, but it's almost 6AM and I'll have to dig into this later.

Nihlus:
Are you happy with the state of this PR?

VelvetThePanda:
I believe we worked out the outstanding issues on Discord, so yes?

PR #275, opened 2023-01-05
Re-type IMessageComponentData.cs#Values to an optional list of strings

carlst99:
The Discord API docs denote that this field should contain a list of `ISelectOption` values. However, in their samples and as identified through testing, it actually always contains a list of strings.

I will open a PR on the Discord API docs about this issue in short order, and link it here once done.


carlst99:
I have now created an associated PR on the Discord API docs - https://github.com/discord/discord-api-docs/pull/5823

PR #277, opened 2023-01-06
Re-type IMessageComponentData.cs#Values as a union array of snowflakes or strings

carlst99:
Following discussion in https://github.com/discord/discord-api-docs/pull/5823, my assumption in #275 appears to be incorrect, and the underlying type of the `values` field depends on the type of select menu used. This PR resolves that; however, given the current implementation works, I would like to leave this as a draft until the API docs PR is resolved with hopefully a final answer.


Remora.Discord.Interactivity/Responders/InteractivityResponder.cs
```diff
@@ -149,13 +148,7 @@ or ComponentType.MentionableSelect
...
         {
             ComponentType.Button => new Dictionary<string, IReadOnlyList<string>>(),
-            ComponentType.StringSelect => Result<Dictionary<string, IReadOnlyList<string>>>.FromSuccess
-            (
-                new Dictionary<string, IReadOnlyList<string>>
-                {
-                    { "values", data.Values.Value }
-                }
-            ),
+            ComponentType.StringSelect when data.Values.Value.IsT1 => BuildParametersFromStringData(data),
```
MazeXP, line 151: StringSelect must be able to handle both union types.
As there might be a `StringSelect` component with only unsigned number options.
Those would be handled kinda invalidly as `Snowflake`. Therefore `StringSelect` must be able to build the parameters for `string` and `Snowflake`.

Remora.Discord.Interactivity/Responders/InteractivityResponder.cs
```diff
@@ -179,6 +172,21 @@ or ComponentType.ChannelSelect
...
+        IMessageComponentData data
+    )
+    {
+        var parameters = new Dictionary<string, IReadOnlyList<string>>();
+
+        if (data.Values.Value.TryPickT1(out var values, out _))
+        {
+            parameters.Add("values", values);
+        }
+
```
MazeXP, line 186: Add handling of `T0` of the `OneOf` due to reason in above comment.

MazeXP (changes requested):
Looks good!
Need to be cautious with unsigned number only options that would cause a `IReadOnlyList<Snowflake>` even for `StringSelect` component.

Remora.Discord.Interactivity/Responders/InteractivityResponder.cs
carlst99: Ah great point, thanks! I've updated the logic correspondingly

Nihlus:
Looks like a fine way to handle it. Please add variant test data for snowflake-typed value lists as well.

Nihlus:
Are there any updates on this PR?

carlst99:
Hiya @Nihlus, no sorry. Discord's response to my docs PR seemed to indicate I hadn't conveyed my request well, and owing to various things at the time I didn't respond.

I could go for round two on the Discord docs, or we could leave things as-is (deserialising to a string) given the indeterminacy of this case, or we could follow Discord's comment that the data type should always match the type of the select menu and proceed with this implementation - happy to follow your call here 🙂.

Nihlus:
This has been on the backburner for a while now - I'm happy to go with this implementation.

carlst99:
Alright. I'll double-check that this implementation is still accurate later today, and mark as ready-for-review.

carlst99:
Testing against Discord indicates this is still accurate 👍

PR #283, opened 2023-01-15
DocFX: Multiple version support

MazeXP:
This PR customizes the build workflow to publish multiple versions of
the documentation.

It is setup to always keep the `main` branch and the latest 5 tags (variable `KEEP_TAG_VERSIONS`).
The `index.html` is edited on every build to redirect to the latest tag.

## Remark
Before merging into `main` branch the `gh-pages` branch must be cleaned up, to consist only of following files:

#### - `index.html`
```html
<html>
  <head>
    <meta http-equiv="refresh" content="0; url=./main" />
  </head>
  <body></body>
</html>
```

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -70,17 +70,19 @@ jobs:
       KEEP_TAG_VERSIONS: 5
 
     steps:
-      - name: Checkout current gh-pages
+      - name: Checkout triggering branch 
+        uses: actions/checkout@v3
+        with:
+          path: base
+
+      - name: Checkout gh-pages
         uses: actions/checkout@v3
         with:
           ref: gh-pages
           path: site
 
-      - name: Clear version
-        working-directory: site
-        run: |
-          rm -rf ${GITHUB_REF_NAME}
-          echo "Cleared ${GITHUB_REF_NAME}"
+      - name: Prepare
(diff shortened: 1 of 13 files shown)


.github/workflows/dotnet.yml
```diff
@@ -34,34 +39,89 @@ jobs:
...
         with:
           name: docfx-site
-          path: site
+          path: site/${{ github.ref_name }}
+
+      - name: Manage versions
+        shell: bash
+        working-directory: site
+        run: |
+          # Read versions
```
Nihlus, line 95: This might be better suited as a separate script file.

MazeXP: Any recommended place?
How about docfx/scripts or .github/scripts?
(Would have to rework the job to also checkout the branch in the `docs` job)

Nihlus: docfx/scripts sounds good to me.

MazeXP:
> Is there a way to avoid having to do the manual work on the gh-pages branch? Having it sort out the details itself would be far preferable.

If me move the part to a script it would also be possible to have the index.html and .nojekyll as a template inside an `docfx/templates` directory.
In addition we could have an aggressive cleanup script that does every file in the root and every folder that does not match either `main` or the pattern `YYYY.XX`.

Nihlus:
Sounds good!

PR #286, opened 2023-03-11
Implement a service that makes permission computation less of a hassle

Nihlus:
This PR adds a new service to Remora.Discord.Extensions that should make member permission calculation more palatable. In particular, it abstracts away the required API calls to gather the neccesary information and exposes a set of methods that can be used when an operation context is available instead of having to pass in contextual values.

Opinions and suggestions are welcome.

### DIFF
--- a/Remora.Discord.Extensions/Services/IPermissionComputationService.cs
+++ b/Remora.Discord.Extensions/Services/IPermissionComputationService.cs
@@ -1,0 +1,126 @@
+//
+//  IPermissionComputationService.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+using System.Threading;
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+using Remora.Discord.API.Abstractions.Objects;
+using Remora.Discord.Commands.Contexts;
+using Remora.Rest.Core;
+using Remora.Results;
+
+namespace Remora.Discord.Extensions.Services;
+
+/// <summary>
+/// Computes permission sets based on available data.
+/// </summary>
+[PublicAPI]
+public interface IPermissionComputationService
+{
+    /// <summary>
+    /// Computes the effective permissions for the user with the given ID.
+    /// </summary>
+    /// <remarks>
+    /// <paramref name="channelID"/> controls whether permission overwrite in any particular channel are taken into
+    /// account. If it is omitted, the computed permissions are valid anywhere in the server where a permission
+    /// overwrite does not otherwise modify the permissions.
+    ///
+    /// If you want to compute effective permissions for some kind of user action, it's likely that you want to set this
+    /// parameter to an appropriate value.
+    /// </remarks>
+    /// <param name="userID">The ID of the user.</param>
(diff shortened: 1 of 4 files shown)


Remora.Discord.Extensions/Services/IPermissionComputationService.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// Computes the effective permissions for the user with the given ID in the context of the current operation.
+    /// </summary>
+    /// <remarks>
+    /// This method uses the guild and channel ID from an available <see cref="IOperationContext"/>.
+    /// </remarks>
+    /// <param name="userID">The ID of the user.</param>
+    /// <param name="inChannel">Whether the computation should use the channel ID from the context.</param>
+    /// <param name="ct">The cancellation token for this operation.</param>
+    /// <exception cref="InvalidOperationError">
+    /// Thrown if no context is available or it does not have the required information.
```
VelvetThePanda, line 73: \>Error
\>Thrown

Did you mean returned? Or Exception?

Remora.Discord.Extensions/Services/PermissionComputationService.cs
```diff
@@ -0,0 +1,198 @@
...
+        Snowflake userID,
+        Snowflake guildID,
+        Snowflake? channelID = null,
+        CancellationToken ct = default
+    )
+    {
+        var getMember = await _guildAPI.GetGuildMemberAsync(guildID, userID, ct);
+        if (!getMember.IsDefined(out var member))
+        {
+            return Result<IDiscordPermissionSet>.FromError(getMember);
```
VelvetThePanda, line 75: Doesn't `FromError(Result)` wrap? Curious because this could potentially be a little problematic depending on how the result is handled (e.g. being thrown up the call stack, where there's already issues with result wrapping as-is)

Remora.Discord.Extensions/Services/IPermissionComputationService.cs
```diff
@@ -0,0 +1,126 @@
...
+using Remora.Rest.Core;
+using Remora.Results;
+
+namespace Remora.Discord.Extensions.Services;
+
+/// <summary>
+/// Computes permission sets based on available data.
+/// </summary>
+[PublicAPI]
+public interface IPermissionComputationService
```
VelvetThePanda, line 37: Should some disclaimer be made about how stale data (from caching) may produce unexpected results?

Remora.Discord.Extensions/Services/PermissionComputationService.cs
```diff
@@ -0,0 +1,198 @@
...
+    public Task<Result<IDiscordPermissionSet>> ComputeContextualMemberPermissions
+    (
+        Snowflake userID,
+        bool inChannel = true,
+        CancellationToken ct = default
+    )
+    {
+        if (_contextInjection.Context is null)
+        {
+            throw new InvalidOperationException("Contextual permission computation requires an available context.");
```
VelvetThePanda, line 116: re: Prior comment; documentation needs to be updated, or perhaps return an error?

VelvetThePanda (approved):
LGTM, see review notes

Remora.Discord.Extensions/Services/PermissionComputationService.cs
Nihlus: In this case, an exception was the intended action. Using the contextual methods in a non-contextual context (hoo boy, words) is programmer error, not data error.

Remora.Discord.Extensions/Services/PermissionComputationService.cs
Nihlus: It does wrap, but there have been changes made to how wrapping actually functions so it shouldn't be as much of an issue as it previously was. `FromError` uses the error set in the passed result directly, so you should always have the nearest concrete error object on the result object you get.

I think you may be thinking of RestError and its related problems, which have some other considerations (but should be far more readable and useful in the latest release of Remora.Rest).

On top of that, FromError is required since the result types differ :P

VelvetThePanda: I was actually thinking of Remora.Discord.Commands and the issue of four-deep wrapping

VelvetThePanda: Furthermore doesn't `FromError` accept an `IResultError` that returns a failed result instead of a wrapped failed result?

Nihlus: There are several overloads (of which one does take `T where T : IResultError`), but in this case it's using this overload <https://github.com/Remora/Remora.Results/blob/main/Remora.Results/Result.cs#L148>, which doesn't cause any nesting in the previously problematic way.

PR #289, opened 2023-03-19
Implement token-scoped cache keys

### LINKED ISSUE #257: [Bug]: Different tokens share the cache
### Description

When using Remora.Discord with different Bearer tokens then the cache 
will be shared among all tokens.

### Steps to Reproduce

1. Switch token used during execution

### Expected Behavior

The cache should differ between token.

### Current Behavior

Cache does not respect token at all.

### Library / Runtime Information

.

Nihlus:
This PR implements simple scoping for the cache library in order to keep data belonging to different tokens separate.

In its current state, it is not a complete implementation since tokens provided via customizations onto single REST requests will not be picked up, and that mechanism may need some more in-depth changes. Consider this a design suggestion out for comments.

Fixes #257.

### DIFF
--- a/Backend/Remora.Discord.Caching.Redis/Services/RedisCacheProvider.cs
+++ b/Backend/Remora.Discord.Caching.Redis/Services/RedisCacheProvider.cs
@@ -21,6 +21,8 @@
 //
 
 using System;
+using System.Security.Cryptography;
+using System.Text;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


VelvetToroyashi (approved):
Needs to be updated for 'recent' Remora changes to caching, but looks good
<sub>Also unsure how this has gone without review up to this point</sub>

Nihlus:
Closing this as stale. There hasn't been much interest in multi-token bots recently.

Issue #254, opened 2022-11-19
`DiscordGatewayClient` allocation and method improvements


VelvetThePanda:
### Description

As it stands right now, the gateway client is not only the single biggest, single-purpose class* in the library, but it is also the top allocator as shown by a profiler. It's unlikely that these allocations pose a memory leak, but it is partially unnecessary when it's possible to reduce these allocations.

Allocations were initially my main goal when it came to sprucing up the client, as described below, but as I continued to look for room for improvement I realized just how much of a mess the client really is (and, I don't fault Jax here, as the client is somewhat non-trivial in its operation).

However, there is a lot of code that could be simplified by being broken out, as I'll describe later in this issue.

* This is excluding `____API` classes, which I would personally consider "multi-purpose". 

### Why This is Needed

As described above, the gateway client is responsible for the top five allocations across the entire span of the application lifetime, even in comparison to EF Core and the available Redis caching, which marshals objects when deserializing. 

The screenshot below is the measurement over a six-hour period, and while the overall allocations are rather trivial for most, and for the most part will be GC'd, these can still be potentially reduced by trivial means. 

![image](https://user-images.githubusercontent.com/42438262/202833530-77797d00-3405-4a40-b6b4-36a4b03753f7.png)

### Alternatives Considered

Short of rewriting the entire client (e.g. #176), there aren't many alternatives to be had here.

### Additional Details

Here is the meat of the document, and I'll point out everything that I think can be reasonably fixed. 

## Passing around `heartbeat_interval`
Call this a nitpick, but I feel like this issue is just large enough to be somewhat of an issue; not necessarily for maintainability or functionality, but for ease of writing code. There's also potential defensive copying at play here, though that's a bit of a micro-optimization.

https://github.com/Remora/Remora.Discord/blob/7935da197307034c5c8ffc8344cb766e40dcaa45/Backend/Remora.Discord.Gateway/DiscordGatewayClient.cs#L585

---

#### Heartbeating should be its own loop

As shown below, the gateway prioritizes heartbeats over user-submitted payloads which is *fine*, but I think this can be slightly improved by adjusting how payload rate-limiting is accounted for. 

Instead of having heartbeating and user payloads both be rate-limiting, we could instead reserve a single slot (heartbeating is ~45s, and it's 120 events/min, so there's only ever 1 heartbeat per ratelimit window).

[...]

PR #300, opened 2023-04-16
Optimize receving payloads¹

VelvetThePanda:
¹This PR only optimizes `WebSocketPayloadTransportService`'s `ReceiveAsync` method.

This PR is one of many that will tackle various issues mentioned in #254.

This one in particular relates to the performance (speed, allocations) concerns raised by the current implementation of receiving a payload (as implemented by the aforementioned class).

Notable changes are: 
- Using `CommunityToolKit.HighPerformance`
- Switching `WebSocketReceiveResult` to its struct-based alternative
- Using reusable² buffers
- Using `JsonSerializer.Deserialize<T>` instead of its async alternative

## Considerations and justifications

It should be considered that as a critical part of the library, it's important that this code is battle-tested, and not just validated by unit-tests. In theory, the logic of the receive flow has not changed, but there's always the possibility for bugs and edge-cases that may be directly or indirectly found becuase of this PR; any bugs reported in regards to receiving gateway payloads should probably reference this PR for ease of tracking.

²The implementation of `ArrayPoolBufferWriter<T>` is backed by `ArrayPool<T>.Shared` by default

As for *why* these changes were made:

`CommunityToolKit.HighPerformance` is, as it says on the tin, a high-performance set of tools and APIs (such as `ArrayPoolBufferWriter<T>`) which could have even futher utlization within Remora after some careful consideration and triaging of potential optimizations (such as those commented upon in #254). (SendAsync was not changed as incoming payloads outnumber outgoing by upward of several orders of magnitude for larger bots).

The change to `ValueWebSocketReceiveResult` is twofold; `ArrayPoolBufferWriter<T>` exposes `Memory<T>` and `Span<T>`, the former of which is accepted as an overload to `ClientWebSocket#ReceiveAsync`, which is returns a struct result instead. This struct only contains two fields, so is much smaller, and never touches the heap. This optimization, if it can be called such, comes for free as a side-effect of other changes.

`JsonSerializer.DeserializeAsync<T>` was changed to its synchronous alternative because:

- JSON Deserialization is synchronous
- The Async overload only reads from the stream asynchronously

```cs
while (true)
{
    bufferState = await bufferState.ReadFromStreamAsync(utf8Json, cancellationToken).ConfigureAwait(false);
    TValue? value = ContinueDeserialize<TValue>(ref bufferState, ref jsonReaderState, ref readStack, jsonTypeInfo);

    if (bufferState.IsFinalBlock)
    {
        return value;
    }
}
```

The latter is especially important to consider, given *where* the data is coming from.

Becuase the only asynchronous part of the deserialization part is reading from the stream, this is rather bad becuase the stream it's reading from is a `MemoryStream`, which means all the data it's reading is already sitting in memory, not on the wire/on disk. Becuase of this, the overhead of async actually causes deserialization to be *slower* than if the synchronous version was used.

### DIFF
--- a/Backend/Remora.Discord.Gateway/Remora.Discord.Gateway.csproj
+++ b/Backend/Remora.Discord.Gateway/Remora.Discord.Gateway.csproj
@@ -14,6 +14,7 @@
     </ItemGroup>
 
     <ItemGroup>
+      <PackageReference Include="CommunityToolkit.HighPerformance" />
       <PackageReference Include="Remora.Extensions.Options.Immutable" />
       <PackageReference Include="System.Threading.Channels" />
     </ItemGroup>
--- a/Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
+++ b/Backend/Remora.Discord.Gateway/Transport/WebSocketPayloadTransportService.cs
@@ -30,6 +30,8 @@ using System.Text.Json;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using CommunityToolkit.HighPerformance;
+using CommunityToolkit.HighPerformance.Buffers;
 using JetBrains.Annotations;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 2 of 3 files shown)


VelvetThePanda:
Deserialization alone was a ~3X increase (w.r.t itself), as seen here:

```md
|                      Method |       Mean |    Error |   StdDev |     Median | Allocated |
|---------------------------- |-----------:|---------:|---------:|-----------:|----------:|
|  SynchronousDeserialization |   391.6 ns |  7.78 ns | 11.88 ns |   390.5 ns |      33 B |
| AsynchronousDeserialization | 1,328.1 ns | 31.64 ns | 92.29 ns | 1,289.9 ns |      97 B |
```

PR #308, opened 2023-07-23
Change thread_id from Form to Query String param

b-rad15:
https://discord.com/developers/docs/resources/webhook#execute-webhook has this listed as a query string parameter not a form parameter and therefore posting to an existing forum thread is broken. I think it also updates the test correctly to account for this but not as sure about that

### DIFF
--- a/Tests/Remora.Discord.Rest.Tests/API/Webhooks/DiscordRestWebhookAPITests.cs
+++ b/Tests/Remora.Discord.Rest.Tests/API/Webhooks/DiscordRestWebhookAPITests.cs
@@ -817,6 +817,7 @@ public class DiscordRestWebhookAPITests
                 b => b
                     .Expect(HttpMethod.Post, $"{Constants.BaseURL}webhooks/{webhookId}/{token}")
                     .WithQueryString("wait", shouldWait.ToString())
+                    .WithQueryString("thread_id", threadID.ToString())
                     .WithJson
                     (
                         j => j.IsObject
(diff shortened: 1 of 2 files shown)


VelvetToroyashi:
`thread_name` is not accounted for (as per the documentation), however I can open a separate PR for that.

VelvetToroyashi (approved):

Tests/Remora.Discord.Rest.Tests/API/Webhooks/DiscordRestWebhookAPITests.cs
```diff
@@ -817,6 +817,7 @@ public async Task PerformsInThreadRequestCorrectly()
                 b => b
                     .Expect(HttpMethod.Post, $"{Constants.BaseURL}webhooks/{webhookId}/{token}")
                     .WithQueryString("wait", shouldWait.ToString())
+                    .WithQueryString("thread_id", threadID.ToString())
```
Nihlus, line 820: Can you double-check that both parameters are detected with a query string check like this?

Nihlus: Yeah, having an exact matcher would be good - can you update this PR and open an issue?

b-rad15: Ok, pushed that change for this code here and submitted the issue as #309

PR #310, opened 2023-08-19
Allow fetching tokens asynchrnously

Itamaram:
In its current form, the SDK expects the token to be fetched synchronously, which might not be a problem when the token is available locally, but if it is to be fetched from a remote source (such as a secret store), leveraging asychronous patterns is best practice, but is not achievable.

This PR changes the `ITokenStore` abstraction to allow support for tokens which require a `Task` to retrieve. Instead of the stored token being of type `string`, instead it is of `ValueTask<string>`. The existing implementation of this interface was kept functionally equivalent, receiving a string as a constructor argument and returning it when the token is requested.
Consumers of the library can implement their own implementations, using different data sources, and choosing their caching logic of choice.

The extension methods `AddDiscordGateway` and `AddDiscordRest` no longer register an implementation of the token store. It is expected that the caller would do that independently. However, backwards compatible overloads are provided that register the default implementation.


VelvetToroyashi:
Changes look good, and I can see this being useful for a variety of applications. At a glance, as far as I can tell, callers must explicitly omit the token to opt into a custom store implementation?

Itamaram:
> Changes look good, and I can see this being useful for a variety of applications. At a glance, as far as I can tell, callers must explicitly omit the token to opt into a custom store implementation?

That's the expectation, yes. The consumer would use either the current implementation:
```cs
serviceCollection
  .AddDiscordRest(ReadTokenFromConfig)
  ;
```
or
```cs
serviceCollection
  .AddSingleton<IAsyncTokenStore, MyCustomTokenStore>
  .AddDiscordRest()
  ;
```

Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -58,12 +58,27 @@ public static IServiceCollection AddDiscordGateway
         Action<IHttpClientBuilder>? buildClient = null
     )
     {
-        serviceCollection
-            .AddDiscordRest
-            (
-                s => (tokenFactory(s), DiscordTokenType.Bot),
-                buildClient
-            );
+        serviceCollection.AddSingleton<IAsyncTokenStore>(ctx => new StaticTokenStore(
```
Nihlus, line 61: Formatting. If you need to break lines, use allman style (same as with braces) for parens as well.

Backend/Remora.Discord.Rest/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -68,6 +68,28 @@ public static IServiceCollection AddDiscordRest
         Func<IServiceProvider, (string Token, DiscordTokenType TokenType)> tokenFactory,
         Action<IHttpClientBuilder>? buildClient = null
     )
+    {
+        serviceCollection.AddSingleton<IAsyncTokenStore>(
```
Nihlus, line 72: Formatting.

Backend/Remora.Discord.Rest/StaticTokenStore.cs
```diff
@@ -20,30 +20,34 @@
...
 
 /// <summary>
-/// Represents a storage class for a single token.
+/// Represents a storage class for a static token.
 /// </summary>
 [PublicAPI]
-public class TokenStore : ITokenStore
+public class StaticTokenStore : IAsyncTokenStore
 {
+    private readonly ValueTask<string> _token;
```
Nihlus, line 35: `ValueTask` instances should generally only be awaited once, since its underlying semantics are different from a normal task. See https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#valid-consumption-patterns-for-valuetasks for an explanation.

A simple fix would be to just store the string and create a new ValueTask on each invocation.

Nihlus (changes requested):

Backend/Remora.Discord.Rest/Extensions/ServiceCollectionExtensions.cs
Itamaram: Which would be the preferred format in this case?
```cs
        serviceCollection.AddSingleton<IAsyncTokenStore>(ctx =>
        {
            var (token, type) = tokenFactory(ctx);
            return new StaticTokenStore(token, type);
        });
````
or
```cs
        serviceCollection.AddSingleton<IAsyncTokenStore>
        (
            ctx =>
            {
                var (token, type) = tokenFactory(ctx);
                return new StaticTokenStore(token, type);
            }
        );
```

Nihlus: The former is fine in single-argument contexts 🙂

Backend/Remora.Discord.Gateway/Extensions/ServiceCollectionExtensions.cs
Itamaram: Formatted appropriately

Nihlus (approved):

Issue #313, opened 2023-09-05
[Bug]: ExceptionErorr while deserializing json for webhook message after passing wait=true parameter


b-rad15:
### Description

While deserializing the author portion of a webhook message response, there is an error that `GlobaName` was expected but not found: "ExceptionError { Message = The data property "GlobalName" did not have a corresponding value in the JSON., Exception = System.Text.Json.JsonException: The data property "GlobalName" did not have a corresponding value in the JSON."

### Steps to Reproduce

Here is the exact test code I was using in xunit to debug this:
```csharp
await SetupDiscordService();
IConfigurationRoot configuration = new ConfigurationBuilder()
    .AddJsonFile("appsettings.json")
    .Build();
//Extract Webhook Params
var webhookUrl = configuration["TestWebhookChannel"];
Assert.NotNull(webhookUrl);
var webhookRegexMatch = RegexData.WebhookRegex().Match(webhookUrl);
var webhookToken = webhookRegexMatch.Groups["Token"].Value;
var webhookIdString = webhookRegexMatch.Groups["Id"].Value;
Snowflake webhookId = new(ulong.Parse(webhookIdString));
Assert.NotNull(DiscordHostService);
//Send The Message
var webhookService = DiscordHostService.Services.GetRequiredService<IDiscordRestWebhookAPI>();
var webhookSendResult =
    await webhookService.ExecuteWebhookAsync(webhookId, webhookToken, true, $"Test Message at {DateTime.Now}", $"Test Username at {DateTime.Now}");
//Check the result
if (!webhookSendResult.IsSuccess)
{
    _testOutputHelper.WriteLine($"Failed to send webhook message: {webhookSendResult.Error}");
    Assert.True(webhookSendResult.IsSuccess);
}
```

### Expected Behavior

GlobalName should be an `Optional<string?>` value of an `IUser` object despite Discord incorrectly marking this as a required nullable string in their API docs: https://discord.com/developers/docs/resources/user#user-object

### Current Behavior

Discord responds with the following after waiting for a webhook message (Ids redacted just in case):
```json
"author": {
    "id": "redacted",
    "username": "test username postman",
    "avatar": null,
    "discriminator": "0000",
    "public_flags": 0,
    "flags": 0,
    "bot": true
    },
```
IUser has a GlobalName property that is string? meaning that when global_name is not present on the author response the json deserializer throws an exception.

### Library / Runtime Information

framework: net7.0
Remora.Discord: 2023.4.0
[...]

PR #316, opened 2023-09-15
Remove global_name from User objects to comply with OpenAPI spec UserPIIResponse and UserResponse objects

b-rad15:
The spec as defined by https://github.com/discord/discord-api-spec/blob/main/specs/openapi.json
Resolves issue #313 
Wasn't sure how to mark it as intentionally left out so let me know if I should change that (or if anyone else wants to just fix it go for it). I also have generated patch files that could revert this once discord actually implements global_name fully, though I assume those could also be generate after the fact just from this pr.

### DIFF
--- a/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IPartialUser.cs
+++ b/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IPartialUser.cs
@@ -29,6 +29,7 @@ namespace Remora.Discord.API.Abstractions.Objects;
 /// <summary>
 /// Represents a partial Discord user.
 /// </summary>
+/// <inheritdoc cref="IUser" path="/remarks"/>
 [PublicAPI]
 public interface IPartialUser
 {
@@ -40,9 +41,6 @@ public interface IPartialUser
 
     /// <inheritdoc cref="IUser.Discriminator" />
     Optional<ushort> Discriminator { get; }
-
-    /// <inheritdoc cref="IUser.GlobalName" />
-    // Optional<string?> GlobalName { get; }
 
     /// <inheritdoc cref="IUser.Avatar" />
     Optional<IImageHash?> Avatar { get; }
--- a/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IPartialUser.cs
+++ b/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IPartialUser.cs
@@ -42,7 +42,7 @@ public interface IPartialUser
     Optional<ushort> Discriminator { get; }
 
     /// <inheritdoc cref="IUser.GlobalName" />
-    Optional<string?> GlobalName { get; }
+    // Optional<string?> GlobalName { get; }
 
     /// <inheritdoc cref="IUser.Avatar" />
     Optional<IImageHash?> Avatar { get; }
--- a/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IUser.cs
+++ b/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IUser.cs
@@ -29,6 +29,9 @@ namespace Remora.Discord.API.Abstractions.Objects;
 /// <summary>
 /// Represents a Discord user.
 /// </summary>
+/// <remarks>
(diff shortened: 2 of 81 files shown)


MazeXP:
The PR looks good to me!

Remark: The field is only not provided for webhooks as Discord is treating them not a valid user, not because it is not implemented

Discord docs under the definition of message object:
> The author object follows the structure of the user object, but is only a valid user in the case where the message is generated by a user or bot user. If the message is generated by a webhook, the author object corresponds to the webhook's id, username, and avatar. You can tell if a message is generated by a webhook by checking for the webhook_id on the message object.

Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IPartialUser.cs
```diff
@@ -42,7 +42,7 @@ public interface IPartialUser
     Optional<ushort> Discriminator { get; }
 
     /// <inheritdoc cref="IUser.GlobalName" />
-    Optional<string?> GlobalName { get; }
+    // Optional<string?> GlobalName { get; }
```
Nihlus, line 45: I'm not super keen on leaving code commented out like this - preferably, you'd leave an explanation at the class/interface level documentation instead in a `<remarks>` section.

Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IUser.cs
```diff
@@ -55,7 +55,11 @@ public interface IUser : IPartialUser
     /// <summary>
     /// Gets the user's display name, if it is set. For bots, this is the application name.
     /// </summary>
-    new string? GlobalName { get; }
+    /// <remarks>
+    /// Intentionally not implemented as this field is not present in Discord's OpenAPI specification.
```
Nihlus, line 59: This, for example, is a good text you can move there instead.

Backend/Remora.Discord.API/API/Gateway/Events/Users/UserUpdate.cs
```diff
@@ -35,7 +35,8 @@ public record UserUpdate
     Snowflake ID,
     string Username,
     ushort Discriminator,
-    string? GlobalName,
+    // GlobalName is not present in the OpenAPI spec, only in documentation and is not returned by all endpoints
```
Nihlus, line 38: No need to comment the individual constructor parameters.

Nihlus (changes requested):

Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IUser.cs
b-rad15: Put this on the remarks for IUser and used inheritdoc to copy just that remark to the other classes/interfaces I removed global_name from

Backend/Remora.Discord.API/API/Gateway/Events/Users/UserUpdate.cs
b-rad15: removed these too

Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IPartialUser.cs
b-rad15: removed these and set the remarks with inheritdoc to IUser's remarks

aPinat:
I feel like there should've been a different solution than removing it outright, as this name as a user's display name is commonly used and there is no other way to get it now.

Discord themselves [say](https://github.com/discord/discord-api-spec#spec-conventions) to follow the docs instead of the OpenAPI spec as well, as it is still in preview.

> This is a preview and it may be not correct. If you find discrepancies between the spec and our [docs](https://discord.com/developers/docs), other than the ones mentioned below, let us know, and follow the docs, not the spec.

aPinat:
I feel like there should've been a different solution for #313 than removing it outright, as this name as a user's display name is commonly used and there is no other way to get it now.

Discord themselves [say](https://github.com/discord/discord-api-spec#spec-conventions) to follow the docs instead of the OpenAPI spec as well, as it is still in preview.

> This is a preview and it may be not correct. If you find discrepancies between the spec and our [docs](https://discord.com/developers/docs), other than the ones mentioned below, let us know, and follow the docs, not the spec.

PR #325, opened 2023-12-31
Revert global_name removal

Octol1ttle:
Reverts #316 now that discord/discord-api-spec#30 is [fixed](https://github.com/discord/discord-api-spec/commit/260d102e6a9783cf811a43e1cdcb16e5c12a296c)

### DIFF
--- a/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IUser.cs
+++ b/Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IUser.cs
@@ -29,9 +29,6 @@ namespace Remora.Discord.API.Abstractions.Objects;
 /// <summary>
 /// Represents a Discord user.
 /// </summary>
-/// <remarks>
-/// global_name intentionally not implemented as this field is not present in Discord's OpenAPI specification, despite being documented.
-/// </remarks>
 [PublicAPI]
 public interface IUser : IPartialUser
 {
@@ -54,6 +51,11 @@ public interface IUser : IPartialUser
     /// </summary>
     /// <remarks>Discord no longer uses discriminators, and migrated users will simply have '0' as their discriminator.</remarks>
     new ushort Discriminator { get; }
+
+    /// <summary>
+    /// Gets the user's display name, if it is set. For bots, this is the application name.
+    /// </summary>
+    new string? GlobalName { get; }
 
     /// <summary>
     /// Gets the user's avatar hash.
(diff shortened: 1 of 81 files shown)


Backend/Remora.Discord.API.Abstractions/API/Objects/Users/IUser.cs
```diff
@@ -55,6 +52,11 @@ public interface IUser : IPartialUser
     /// <remarks>Discord no longer uses discriminators, and migrated users will simply have '0' as their discriminator.</remarks>
     new ushort Discriminator { get; }
 
+    /// <summary>
+    /// Gets the user's display name, if it is set. For bots, this is the application name.
+    /// </summary>
+    new string? GlobalName { get; }
```
Nihlus, line 58: This should be an `Optional<string?>` in order to comply with the spec.

Backend/Remora.Discord.API/API/Gateway/Events/Users/UserUpdate.cs
```diff
@@ -35,6 +35,7 @@ public record UserUpdate
     Snowflake ID,
     string Username,
     ushort Discriminator,
+    string? GlobalName,
```
Nihlus, line 38: This should be an `Optional<string?>` in order to comply with the spec.

Backend/Remora.Discord.API/API/Objects/Users/User.cs
```diff
@@ -36,6 +36,7 @@ public record User
     Snowflake ID,
     string Username,
     ushort Discriminator,
+    string? GlobalName,
```
Nihlus, line 39: This should be an `Optional<string?>` in order to comply with the spec.

Backend/Remora.Discord.API/API/Objects/Users/UserMention.cs
```diff
@@ -34,6 +34,7 @@ public record UserMention
     Snowflake ID,
     string Username,
     ushort Discriminator,
+    string? GlobalName,
```
Nihlus, line 37: This should be an `Optional<string?>` in order to comply with the spec.

Tests/Remora.Discord.Tests/Samples/Objects/USER/USER.optionals.json
```diff
@@ -2,5 +2,6 @@
   "id": "999999999999999999",
   "username": "none",
   "discriminator": "9999",
+  "global_name": "none",
```
Nihlus, line 5: This (along with the rest of the sample files) need to be updated in order to comply with the spec. This variant should not have `global_name` in the payload.

All other sample files (i.e, the ones that don't define the property in their type) can have the property removed as well, since the object used when in a nested property should be the minimal valid structure (i.e, the one in XX.optionals.json). 

You should still retain them in UserMention and the gateway events, though, since they directly define/use the property.

Nihlus (changes requested):

PR #339, opened 2024-08-17
Add invite metadata to guild invites response

starshine-sys:
Adds the keys in the invite metadata object (https://discord.com/developers/docs/resources/invite#invite-metadata-object) to the type returned by GET /guilds/{guildID}/invites.

### DIFF
--- a/Backend/Remora.Discord.API/API/Objects/Invites/InviteWithMetadata.cs
+++ b/Backend/Remora.Discord.API/API/Objects/Invites/InviteWithMetadata.cs
@@ -34,7 +34,7 @@ public record InviteWithMetadata
     IPartialChannel? Channel,
     int Uses,
     int MaxUses,
-    int MaxAge,
+    TimeSpan MaxAge,
     bool IsTemporary,
     DateTimeOffset CreatedAt,
     Optional<IUser> Inviter = default,
--- a/Backend/Remora.Discord.API/API/Objects/Invites/InviteWithMetadata.cs
+++ b/Backend/Remora.Discord.API/API/Objects/Invites/InviteWithMetadata.cs
@@ -1,0 +1,48 @@
+//
+//  InviteWithMetadata.cs
+//
+//  Author:
+//       Jarl Gullberg <jarl.gullberg@gmail.com>
+//
+//  Copyright (c) Jarl Gullberg
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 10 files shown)


Backend/Remora.Discord.API/API/Objects/Invites/InviteWithMetadata.cs
```diff
@@ -0,0 +1,48 @@
...
+/// <inheritdoc cref="IInviteWithMetadata" />
+public record InviteWithMetadata
+(
+    string Code,
+    Optional<IPartialGuild> Guild,
+    IPartialChannel? Channel,
+    int Uses,
+    int MaxUses,
+    int MaxAge,
+    bool IsTemporary,
```
VelvetToroyashi, line 38: This is missing the json property name configuration (should be `temporary` in json, as per the docs)

VelvetToroyashi (changes requested):
Thanks for this PR! Overall looks good - see below comments.

Also tests should be updated to reflect the new data/return type

MazeXP:
Thanks for the PR. Velvet mentioned it already that you are missing the JSON property name configuration for the `temporary` field and the registration of the type itself in `Remora.Discord.API`.
Add the type [here](https://github.com/Remora/Remora.Discord/blob/main/Backend/Remora.Discord.API/Extensions/ServiceCollectionExtensions.cs#L783).
For the property configuration have a look [here]( https://github.com/Remora/Remora.Discord/blob/main/Backend/Remora.Discord.API/Extensions/ServiceCollectionExtensions.cs#L336-L338).

Jax must decide if the name `InviteWithMetadata` is fine or we should stick to the original name `InviteMetadata` as it is called in the Discord docs.

For the test files you would need to create a subfolder in `Remora.Discord.Tests`: [here](https://github.com/Remora/Remora.Discord/tree/main/Tests/Remora.Discord.Tests/Samples/Objects)
and for the actual test you need to add an additional test class in `Remora.Discord.API.Tests`: [here](https://github.com/Remora/Remora.Discord/blob/main/Tests/Remora.Discord.API.Tests/API/Objects/Invites/)

starshine-sys:
I've registered the type now, thanks for the pointers!  
I only named the type `InviteWithMetadata` because it's not actually a *separate* object, but an extension of the `Invite` type, and I couldn't quickly find a similar situation in the existing library--either name is fine.

There's already a test for this endpoint [here](https://github.com/Remora/Remora.Discord/blob/5ada687d1e6d40b37d4732acd535721ad25c76cf/Tests/Remora.Discord.Rest.Tests/API/Guild/DiscordRestGuildAPITests.cs#L2690), but it just tests with an empty array response. I couldn't find any test for an endpoint that returns an array that *doesn't* use an empty array response, so I'm not 100% sure how to add a test for that, sorry.

MazeXP:
The endpoint does not need more testing. It is a test about the actual JSON object.

starshine-sys:
> The endpoint does not need more testing. It is a test about the actual JSON object.

Ah, got it! I didn't realize those tests were separate. I added those now, and they all pass (and aren't skipped), so I *think* I did it right?

MazeXP:
LGTM! Thanks again. Now let's wait on review of @Nihlus and his final word about the `InviteMetadata` vs `InviteWithMetadata`.

PR #344, opened 2024-10-28
Message updates are now full messages .

VelvetToroyashi:
This commit changes the IMessageUpdate interface (and subsequently, its implementation)to implement the IMessageInterface instead of the IPartialMessage interface.

This change is due to Discord returning a full message object on MESSAGE_UPDATE.

BREAKING-CHANGE: IMessageUpdate now implements IMessage (which implements IPartialMessage),
`Mentions` is no longer optional. The ad-hoc implementation for IPartialMessage was kept out of convenience (and to avoid further breaking changes)


Backend/Remora.Discord.API.Abstractions/API/Gateway/Events/Messages/IMessageUpdate.cs
```diff
@@ -31,7 +31,7 @@ namespace Remora.Discord.API.Abstractions.Gateway.Events;
 /// Represents the creation of a message.
```
Octol1ttle, line 31: unrelated to PR but "creation"?

Backend/Remora.Discord.API.Abstractions/API/Gateway/Events/Messages/IMessageUpdate.cs
```diff
@@ -48,8 +48,11 @@ public interface IMessageUpdate : IPartialMessage, IGatewayEvent
     /// <summary>
     /// Gets a list of users mentioned in the message.
     /// </summary>
-    new Optional<IReadOnlyList<IUserMention>> Mentions { get; }
+    new IReadOnlyList<IUserMention> Mentions { get; }
```
Nihlus, line 51: We should be able to drop this (and the explicit implementations below) outright as the forwarding implementation for IPartialMessage is already handled by IMessage. We're already breaking the interface, so we might as well clean things up.

Nihlus (changes requested):

PR #367, opened 2026-07-12
fix: add cancellation token to CacheService.TryGetPreviousValueAsync

Octol1ttle:
not breaking


Backend/Remora.Discord.Caching/Services/CacheService.cs
```diff
@@ -104,10 +104,11 @@ public ValueTask<Result<TInstance>> TryGetValueAsync<TInstance>(CacheKey key, Ca
     /// Attempts to retrieve the previous value of the given key from the eviction cache.
     /// </summary>
     /// <param name="key">The cache key.</param>
+    /// <param name="ct">A cancellation token to cancel the operation.</param>
     /// <typeparam name="TInstance">The instance type.</typeparam>
     /// <returns>A <see cref="Result"/> that may or not have succeeded.</returns>
-    public ValueTask<Result<TInstance>> TryGetPreviousValueAsync<TInstance>(CacheKey key)
-        where TInstance : class => _cacheProvider.RetrieveAsync<TInstance>(new KeyHelpers.EvictionCacheKey(key));
+    public ValueTask<Result<TInstance>> TryGetPreviousValueAsync<TInstance>(CacheKey key, CancellationToken ct = default)
```
Banane9, line 110: Adding a parameter is definitely a binary incompatible change, even if it is code compatible.

Banane9: You'd need to add an overload that doesn't have the parameter, if you want to preserve binary compatibility

Octol1ttle: not sure if binary compatibility is important here, as far as I'm aware "breaking" is used in the context of source-breaking in this library

Banane9: Not sure, that's why this is only a comment and not a yes/no review :D

But in terms of semver, this would be a major version increment to me, as is.

Nihlus: I tend to err on the side of caution with major version changes. Best not to surprise users and version numbers are cheap.

